Issue #7, opened 2016-05-09
Simulate timestamps and/or etags


ndt93:
Each update pass should only request for new changes instead of all addresses. This necessarily requests some change sets marker such as `lastModifiedDate` or `ETags`

ndt93:
Each data item might also need time related fields such as `updatedAt` in order to detect stale data from the cloud for example.

PR #26, opened 2016-05-10
Add updatedAt field to Person and use it to validate data updates #7

ndt93:
This does not complete #7. However, rest of the feature mentioned in #7 can be build on top of this.


ndt93:
This might make it more troublesome to make manual change to the mirror file for debugging (need to update the `updatedAt` field too), so might not be a good idea to merge at the moment.

PR #34, opened 2016-05-11
Support adding/editing/removal of contact groups to an address book #29

### LINKED ISSUE #29: Support adding/editing/removal of contact groups to an address book
For now, groups can only be accessed through the dummy data which has only 2 contact groups: `friends` and `relatives`.

m133225:
Fixes #29.

- [x] adding new contact groups
- [x] editing contact groups
- [x] removing contact groups

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 build/
 .gradle/
 *.iml
+*.xml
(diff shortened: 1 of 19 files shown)


m133225:
This PR does not fix the fact that removal from the local storage does not remove the respective data in the cloud storage.

A few interactions added:
- Double clicking on a group item will launch an edit dialog
- Right clicking on a group item will launch a context menu for adding, editing and removing
- Alternative way of adding a new group quickly using the navigation bar (just to test that the dialogs can be launched anywhere if passed the correct parameters.)

PR #42, opened 2016-05-12
Enforced uniqueness of data items + (Selection is lost during syncing #41)

### LINKED ISSUE #41: Selection is lost during syncing
The selection (dark blue) is lost during syncing.

YijinL:
Fixes #41 

- Loading data now updates the model by diffing instead of full replacement
- New contract: no duplicate Persons or ContactGroups allowed
- Loading from external files detects duplicates
- User actions causing duplicates are also detected

Todo:

- [x] 1. Person equality will use first and last name
- [x] 2. Group equality uses name and checks from `Person.equals`
- [x] 3. Loading data from a mirror or main file with duplicate data will terminate the current action and show an error dialogue
- [x] 4. Loading data from above sources will diff the incoming data with current state and update accordingly (add, delete, update) with minimal changes to the `ObservableList`
- [x] 5. Adding data will show an error dialogue if it results in duplicate data (eg adding Person with same name)
- [x] 6. Editing data will show an error dialogue if it results in duplicate data (eg editing Person name to be same as another existing Person)
- [x] 7. Handle error popup from save file loading problems at startup in mainapp


src/main/java/address/MainApp.java
```diff
@@ -35,15 +41,29 @@ public void start(Stage primaryStage) {
...
+        final File initialFile = PreferencesManager.getInstance().getPersonFile();
+        AddressBookWrapper dataFromFile;
+        Exception fileLoadException = null;
+        try {
+             dataFromFile = StorageManager.loadDataFromSaveFile(initialFile);
+        } catch (JAXBException | FileContainsDuplicatesException e) {
+            fileLoadException = e;
+            dataFromFile = new AddressBookWrapper();
+        }
+
```
damithc: This code seems too low level to be here. this method should not even be aware of JAXBException etc. Push to StorageManager?

YijinL: Right now if the **initial** load detects duplicate data, we will inform the user with an error dialogue (for consistency, the same dialogue as the one triggered when the user tries to **manually** load a file containing duplicates). 
The implementation also uses the same event `FileOpeningExceptionEvent`.

The problem is that the `setupComponents` method has to **wait** till `mainController` is loaded in order to trigger the error dialogue code, and we need to remember the result of the load attempt till `mainController` is ready before posting the appropriate event. I'm not sure how to nicely shift the logic downwards to `StorageManager`, since we have to track both cases 
- [no duplicates => `AddressBookHandler`] 
- [duplicates => event or exception]

src/main/java/address/MainApp.java
```diff
@@ -35,15 +41,29 @@ public void start(Stage primaryStage) {
...
         modelManager = new ModelManager(dataFromFile);
         storageManager = new StorageManager(modelManager);
         mainController = new MainController(this, modelManager, config);
         syncManager = new SyncManager();
         syncManager.startSyncingData(config.updateInterval, config.isSimulateRandomChanges);
+
+        if (fileLoadException != null) {
+            EventManager.getInstance().post(new FileOpeningExceptionEvent(fileLoadException, initialFile));
+        }
```
YijinL: @damithc you can see the other half of the user error feedback logic here. It's separated because of the forced component load order

src/main/java/address/MainApp.java
YijinL: If we change `mainController`'s initialization to be independent of reading data from the file, then everything becomes a lot simpler. (We can use the normal load-from-save-file code without any problems)

damithc: Sure, go ahead and change initialization logic if that makes things better. It has been changed several times already.

YijinL: Or we could just not inform the user of the duplicate problem when it occurs at first load. Then silently substitute sample data or empty data. But it seems wrong.

YijinL:
@yl-coder

yl-coder:
@YijinL Merged #52 .

PR #46, opened 2016-05-13
Implement testing #45

### LINKED ISSUE #45: Implement testing
- both unit test and gui tests
- exploration/research of different technologies/methods of executing them

m133225:
Fixes #45.

Experimented with:
- Automaton VS **TestFX**
  - Automaton: Easy to set-up, slightly better set of testing functionalities, but does not support headless testing
  - TestFX: few issues to figure out about loading of fxml, but can be fixed using a workaround.
- TestNG vs **JUnit**
  - Seems like TestNG is slightly more customisable e.g. re-running failed tests. But it will take time to fully figure out the implementation & may make the project organization messy (many xml files, difficulty implementing in gradle).

Other notes:
- `./gradlew` (without arguments) will run style checks, **headless testing**, and cleaning.
- `./gradlew check clean` will run style checks, **headful testing**, and cleaning.
- `./gradlew headless` will run only the headless tests.  
This difference is probably unintuitive, but there are couple of problems with gradle (conflicting plugins, inability to define combination of tasks aside from defaultTasks) which we will need to figure out.

I have split the tests into `guitests` and `unittests`. And by gradle settings, each test class will restart the JVM (and thus the whole application). This may not be suitable for Mac OS users to run in the background since the restarting of JVM will cause the user to lose window focus. Note that there is a setting under `Run` > `Edit Configurations` > `JUnit` > `Fork mode` which can be set to `class` to make IDEA follow this behaviour. Methods within a test class may thus be dependent on each other, and should be taken note of.

Will merge this PR somewhere next week (16th?) after considering/reading up on more options.

### DIFF
--- a/src/test/java/address/UITest.java
+++ b/src/test/java/address/UITest.java
@@ -1,0 +1,38 @@
+package address;
+
+import javafx.fxml.FXMLLoader;
+import javafx.geometry.Insets;
+import javafx.scene.Group;
+import javafx.scene.Parent;
+import javafx.scene.control.Alert;
+import javafx.scene.layout.BorderPane;
+import javafx.stage.Stage;
+import org.junit.After;
+import org.junit.Before;
+import org.loadui.testfx.GuiTest;
+import org.testfx.api.FxToolkit;
+
+import java.util.concurrent.TimeoutException;
+
+public class UITest extends GuiTest {
+    Stage stage;
+
+    @Override
+    protected Parent getRootNode() {
+        Parent root = stage.getScene().getRoot();
+        stage.getScene().setRoot(new Group());
+        return root;
+    }
+
+    public UITest() {
+        try {
+            FXMLLoader.setDefaultClassLoader(MainApp.class.getClassLoader()); // workaround to fxml loading problems
+            FxToolkit.registerPrimaryStage();
+            FxToolkit.setupApplication(TestApp.class);
+        } catch (TimeoutException e) {
+            e.printStackTrace();
+        }
+
+        stage = FxToolkit.toolkitContext().getRegisteredStage();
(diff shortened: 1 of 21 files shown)


src/test/java/address/UITest.java
```diff
@@ -0,0 +1,38 @@
...
+    Stage stage;
+
+    @Override
+    protected Parent getRootNode() {
+        Parent root = stage.getScene().getRoot();
+        stage.getScene().setRoot(new Group());
+        return root;
+    }
+
+    public UITest() {
```
damithc: `UiTest`

src/test/java/address/UITest.java
```diff
@@ -0,0 +1,38 @@
...
+import javafx.scene.layout.BorderPane;
+import javafx.stage.Stage;
+import org.junit.After;
+import org.junit.Before;
+import org.loadui.testfx.GuiTest;
+import org.testfx.api.FxToolkit;
+
+import java.util.concurrent.TimeoutException;
+
+public class UITest extends GuiTest {
```
damithc: The child class name should sound more specific than the parent class.

m133225:
Seems like there is also a way to test the UI components individually. This could be used to test UI components that have significant interactions with the user, e.g. the pickers. However, this means that there will be 3 different types of tests
- GUI Tests

damithc:
Sure, can have another package. It's good to hear we can test GUI components separately. :+1:

m133225:
Should be ready to merge. This PR does not include mockito and coveralls.

@damithc Could you help to enable the `HubTurbo/addressbook` repository on [Travis](https://travis-ci.org/profile/HubTurbo)?

Issue #60, opened 2016-05-16
Add a component for handling keyboard shortcuts


damithc:
Try to keep the keyboard shortcuts in one place

YijinL:
Some keyboard shortcuts are declared and embedded in the `fxml` view files. See [RootLayout.fxml](https://github.com/HubTurbo/addressbook/blob/f080b2c27d65b54124292c8f8e9778763a268d5d/src/main/resources/view/RootLayout.fxml#L17-L20) for an example.

Should they be moved out of those files and set programmatically?

damithc:
Yes, I'm planning to move them to one place if that's possible.

PR #62, opened 2016-05-17
Add a component for handling keyboard shortcuts #60

damithc:


damithc:
Still work in progress

damithc:
@m133225 merge if ok

src/main/java/address/events/DeleteRequestEvent.java
```diff
@@ -0,0 +1,19 @@
...
+import address.model.ContactGroup;
+import address.model.Person;
+
+import java.io.File;
+import java.util.List;
+
+/**
+ * Indicates a request for deleting a person
+ */
+public class DeleteRequestEvent {
```
m133225: This class and `EditRequestEvent` don't seem to be used.

src/main/java/address/controller/PersonOverviewController.java
```diff
@@ -147,10 +149,44 @@ private ContextMenu createContextMenu(){
...
         editMenuItem.setOnAction(e -> handleEditPerson());
         MenuItem deleteMenuItem = new MenuItem("Delete");
+        deleteMenuItem.setAccelerator(ShortcutsManager.SHORTCUT_PERSON_DELETE);
         deleteMenuItem.setOnAction(e -> handleDeletePerson());
         contextMenu.getItems().addAll(editMenuItem, deleteMenuItem);
         return contextMenu;
     }
+
+    @Subscribe
+    private void HandleDeleteRequestEvent(DeleteRequestEvent event) {
```
m133225: This and `HandleEditRequestEvent` are unused. Should be named `handleDeleteRequestEvent` (no capitalization of `H`).

src/main/java/address/MainApp.java
```diff
@@ -61,6 +63,10 @@ protected void setupComponents() {
         storageManager = new StorageManager(modelManager);
         mainController = new MainController(this, modelManager, config);
         syncManager = new SyncManager();
+
+        shortcutsManager = new ShortcutsManager();
+        EventManager.getInstance().registerHandler(shortcutsManager);
```
m133225: Should this be placed in the constructor of `ShortcutsManager` instead? To be consistent with the other components which involve event handlers.

m133225:
Aside from minor style issues, I wonder if we should be using events for keyboard shortcuts? Rather, I think they should only be used to pass information from the model to the UI.
I guess we can discuss this later (together with #8).

src/main/java/address/MainApp.java
damithc: I'll move it.

src/main/java/address/events/DeleteRequestEvent.java
damithc: Removing

src/main/java/address/controller/PersonOverviewController.java
damithc: Removing

PR #73, opened 2016-05-17
Sample data not saved #68

### LINKED ISSUE #68: Sample data not saved
Steps:
1. Append sample data
2. 'Save as' data.xml
Problem: appended sample data disappears from list
3. Try to append sample data again. Clash of data reported.
4. When examining the file manually, only label groups are in it (no persons).

YijinL:
Hotfixes #68.

Apparently this bug has been there since the very start.
- The only time we wrote to cloud was when we detected a `LocalModelChangeEvent`
- If we save as a different file, technically there wasnt a local mode change, so the cloud isn't updated and the old existing cloud data overwrites the model on the next sync cycle
- saving as a completely new file meant that the mirror file was empty, so the cloud data (nothing) overwrites the sample data

Solution:
- `SyncManager` gets a new handler for `SaveRequestEvent`, so now save requests trigger writes to both local and cloud files

### DIFF
--- a/src/main/java/address/sync/SyncManager.java
+++ b/src/main/java/address/sync/SyncManager.java
@@ -4,6 +4,7 @@ package address.sync;
 import address.events.EventManager;
 import address.events.LocalModelChangedEvent;
 import address.events.NewMirrorDataEvent;
+import address.events.SaveRequestEvent;
 import address.exceptions.FileContainsDuplicatesException;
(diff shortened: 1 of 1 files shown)


YijinL:
@damithc found the cause, I'll merge the fix now?

YijinL:
Hotfix till #89 is solved

PR #91, opened 2016-05-18
Make save and mirror files independent #89

### LINKED ISSUE #68: Sample data not saved

### LINKED ISSUE #89: Make save and mirror files independent
Preferences will record mirror file path separately. 
_User facing feature changes:_
<hr>
**New File**
Clears save and mirror preferences, clears model. 

<hr>
**Open File**
Sets and opens a user-specified save file, also allows changing of mirror file together as an atomic action.

<hr>
**Save As**
Lets user specify new save file and saves model to file. Mirror remains the same.

<hr>
**NEW: Change Sync Mirror**
Lets the user specify new mirror file for syncing. Save file remains unchanged.

<hr>

YijinL:
Resolves #89, fixes #68.

**Separating mirror and save files**
- [x] Differentiate mirror and save file name changed events
- [x] Show both save and mirror file names on UI
- [x] Add mirror file preference key
- [x] Syncing uses mirror file preference

**User Operations**
- [x] New File
- [x] Open File
- [x] Save As
- [x] Choose Mirror


src/main/java/address/preferences/PrefsManager.java
```diff
@@ -0,0 +1,91 @@
+package address.preferences;
```
damithc: If you name the class `PrefManager`, name the package `pref`, to be consistent with other packages.

src/main/java/address/controller/RootLayoutController.java
```diff
@@ -56,13 +75,50 @@ public void setShortcuts(){
...
+        final FileChooser.ExtensionFilter extFilter = new FileChooser.ExtensionFilter("XML files (*.xml)", "*.xml");
+        final FileChooser fileChooser = new FileChooser();
+
+        fileChooser.getExtensionFilters().add(extFilter);
+        final File currentFile = PrefsManager.getInstance().getSaveLocation();
+        fileChooser.setInitialDirectory(currentFile.getParentFile());
+        return fileChooser;
+    }
+
+
```
YijinL: Just shifting non ui-handler code up where it's more visible.

YijinL:
Current behaviour matches master branch behaviour; 

Should merge before mirror-model-savefile sync contract is decided upon and implementation starts.

src/main/java/address/controller/RootLayoutController.java
```diff
@@ -56,13 +75,50 @@ public void setShortcuts(){
...
     /**
      * Creates a new empty address book at the default filepath
      */
     @FXML
     private void handleNew() {
-        PreferencesManager.getInstance().setPersonFilePath(null);
-        modelManager.resetData(Collections.emptyList(), Collections.emptyList());
+        PrefsManager.getInstance().clearSaveLocation();
+        PrefsManager.getInstance().clearMirrorLocation();
+        modelManager.clearModel();
```
yl-coder: Extract this 3 lines into clearXXX() method?

src/main/java/address/controller/RootLayoutController.java
```diff
@@ -27,21 +27,40 @@
  */
 public class RootLayoutController {
 
+    public static final String SAVE_LOC_TEXT_PREFIX = "Save File: ";
```
m133225: These should be private.

src/main/java/address/prefs/PrefsManager.java
```diff
@@ -0,0 +1,94 @@
...
+
+import address.events.EventManager;
+import address.events.MirrorLocationChangedEvent;
+import address.events.SaveLocationChangedEvent;
+
+import java.util.prefs.Preferences;
+import java.io.File;
+
+/**
+ * Manages saving/retrieving of preferences in the registry.
```
yl-coder: I'm not too sure about the convention for this one. But i think it will be useful for developer to know if the class is a singleton class through the class descriptor.

src/main/java/address/prefs/PrefsManager.java
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * Clears the current preferred save file path.
+     */
+    public void clearSaveLocation() {
+        userPrefs.remove(PREF_NODE_NAME_SAVE_LOC);
+        EventManager.getInstance().post(new SaveLocationChangedEvent(null));
+    }
+
+
```
yl-coder: extra line

YijinL: strange i cant see the line in my ide

YijinL: oh wait intellij hides the last newline, i didnt know

src/main/java/address/sync/CloudSimulator.java
```diff
@@ -35,37 +36,41 @@ public CloudSimulator(boolean shouldSimulateUnreliableNetwork) {
...
+
         } catch (JAXBException e) {
             e.printStackTrace();
             System.out.println("File not found or is not in valid xml format : " + cloudFile);
+
         } catch (InterruptedException e) {
             e.printStackTrace();
         }
-        return modifiedData;
+        return Optional.empty();
```
m133225: I think this shouldn't be empty, but rather the newly modified data due to simulated corruption, which also means that `Optional` may not be suitable here.

src/main/java/address/prefs/PrefsManager.java
YijinL: Added that information in the line below

src/main/java/address/MainApp.java
```diff
@@ -48,18 +51,19 @@ public void init() throws Exception {
 
     @Override
     public void start(Stage primaryStage) {
+        LoggerProvider.setLevel(Level.SEVERE);
         setupComponents();
         mainController.start(primaryStage);
 
-        EventManager.getInstance().post(new LoadDataRequestEvent(PreferencesManager.getInstance().getPersonFile()));
+
+        // initial load (requires mainController to be started.
```
HansNewbie: No closing bracket? Also, can clarify if the double new line above is intentional?

src/main/java/address/MainApp.java
```diff
@@ -48,18 +51,19 @@ public void init() throws Exception {
...
+        EventManager.getInstance().post(new LoadDataRequestEvent(PrefsManager.getInstance().getSaveLocation()));
         syncManager.startSyncingData(config.updateInterval, config.simulateUnreliableNetwork);
     }
 
     protected void setupComponents() {
         config = getConfig();
-        PreferencesManager.setAppTitle(config.appTitle);
-
-        modelManager = new ModelManager(new AddressBook());
+        modelManager = new ModelManager(new AddressBook()); // empty model first
```
HansNewbie: Not sure if the comment is necessary, especially that `MainApp` does not need to know about it.

src/main/java/address/controller/MainController.java
```diff
@@ -60,7 +61,7 @@ public MainController(MainApp mainApp, ModelManager modelManager, Config config)
 
     public void start(Stage primaryStage) {
         this.primaryStage = primaryStage;
-        setTitle(config.appTitle, PreferencesManager.getInstance().getPersonFile());
+        primaryStage.setTitle(config.appTitle);
```
HansNewbie: Just confirming: we don't want to show filename in the title? It might be a useful info.

HansNewbie: Nvm, saw below using root layout.

src/main/java/address/controller/RootLayoutController.java
```diff
@@ -71,9 +127,18 @@ private void handleNew() {
...
+        if (toOpen == null) return;
+        PrefsManager.getInstance().setSaveLocation(toOpen);
+        EventManager.getInstance().post(new LoadDataRequestEvent(toOpen));
+    }
+
+    @FXML
+    private void handleChooseMirror() {
+        // Show open file dialog
+        File toSyncWith = getXmlFileChooser().showOpenDialog(mainController.getPrimaryStage());
+        if (toSyncWith == null) return;
```
HansNewbie: Is it possible to use `Optional`?

src/main/java/address/sync/CloudSimulator.java
YijinL: The newly modified data is returned at the end of the `try` block above; 

this line is only reached if there was a file read exception, which matches the case where an unknown error prevents us from retrieving data from the cloud, so an `Optional.empty()` is returned as a signal that retrieving failed

src/main/java/address/prefs/PrefsManager.java
```diff
@@ -0,0 +1,94 @@
...
+        if (instance == null){
+            instance = new PrefsManager();
+        }
+        return instance;
+    }
+
+    private PrefsManager() {}
+
+    /**
+     * @return the current save file preference or the default temp file if there is no recorded prederence.
```
HansNewbie: typo. `preference`.

src/main/java/address/prefs/PrefsManager.java
```diff
@@ -0,0 +1,94 @@
...
+     */
+    public File getSaveLocation() {
+        final String filePath = userPrefs.get(PREF_NODE_NAME_SAVE_LOC, null);
+        return filePath == null ? new File(DEFAULT_TEMP_FILE_PATH) : new File(filePath);
```
HansNewbie: Use `Optional`? Or we can skip using `filePath` altogether.

src/main/java/address/controller/RootLayoutController.java
YijinL: The `FileChooser.showOpenDialog` API returns `null` if the file choosing action was cancelled, we will have to catch the `null` as close to the call as possible.

I will change the `getXmlFileChooser` method to `getFileChosenByUser` or something like that and catch the null there instead.

YijinL: Hmm but doing that means we cannot access the file chooser and do things like set the starting directory. I think we can just catch `null` here, since it's caused by external libraries. Or is there a better way?

src/main/java/address/prefs/PrefsManager.java
YijinL: I offered to use optional, but prof preferred returning a valid default file

YijinL: `filePath` is used in 2 locations on line 38, inlining it would make the ternary statement very long

src/main/java/address/controller/RootLayoutController.java
HansNewbie: Let's keep it this way then.

src/main/java/address/controller/RootLayoutController.java
YijinL: It seems very readable to me, and there is nowhere else that reuses this code right now.

Can't come up with a good name for the extracted method :(
Need to find a reasonable abstraction that covers these 3 steps for the name.

src/main/java/address/prefs/PrefsManager.java
HansNewbie: Right, I missed that. Sure, keep it this way.

HansNewbie:
Just wondering. `SyncManager` no longer handle `SaveRequestEvent`. From the architecture diagram, `SyncManager` ultimately only deals with model. I assume that's the reason for removing the event handler.

If that is indeed the case, maybe we need to change the header comment of the class from `Syncs data between a mirror file and the primary data file` to something more appropriate. At the same time, it should have no knowledge of primary data file in which we should not mention in in [line 44](https://github.com/HubTurbo/addressbook/pull/91/files#diff-35e8f52547709ab2a599a63ec8b2b1eaR44)

YijinL:
Yeah, I brought this up with prof too. He has a different idea about how saving should work (saving should not overwrite the mirror, the sync component should handle it based on timestamps).

Also that event handler was added as part of a hotfix #73

src/main/java/address/sync/CloudSimulator.java
m133225: I guess this will do for now, until we finalise the cloud APIs.

YijinL:
If everything is ok I'll merge this? There is very little overlap with the other currently active pull requests. 
If there are no objections by tomorrow morning I will just merge it so I can work on my next PR that is dependent on the changes here

src/main/java/address/MainApp.java
```diff
@@ -2,14 +2,15 @@
 
 import address.model.AddressBook;
 import com.teamdev.jxbrowser.chromium.BrowserCore;
+import com.teamdev.jxbrowser.chromium.LoggerProvider;
```
damithc: At some point we should prly our own logger instead of jxbrowser one?

src/main/java/address/controller/RootLayoutController.java
```diff
@@ -56,13 +75,50 @@ public void setShortcuts(){
...
+    private void handleSaveLocationChangedEvent(SaveLocationChangedEvent e) {
+        updateSaveLocDisplay();
+    }
+
+    @Subscribe
+    private void handleMirrorLocationChangedEvent(MirrorLocationChangedEvent e) {
+        updateMirrorLocDisplay();
+    }
+
+    private void updateSaveLocDisplay() {
```
damithc: `Loc` and `Location` both used in method names. Standardize on one or the other? The longer one is preferred.

src/main/java/address/MainApp.java
YijinL: This is to control the jxbrowser logging output, I only used it at the app startup to squash all non-severe messages (see line 54).

`LoggerProvider.setLevel(Level.SEVERE);`

It's not directly related to this PR, should I remove it and push it as a separate commit to master?

src/main/java/address/storage/StorageManager.java
```diff
@@ -34,17 +36,17 @@ private void handleLoadDataRequestEvent(LoadDataRequestEvent ofe) {
...
 
     @Subscribe
-    private void handleLocalModelSyncedEvent(LocalModelSyncedFromCloudEvent lmse) {
-        final File targetFile = PreferencesManager.getInstance().getPersonFile();
+    private void handleLocalModelSyncedEvent(LocalModelSyncedFromCloudEvent e) {
         System.out.println("Local data synced, saving to primary data file");
-        saveDataToFile(targetFile, lmse.personData, lmse.groupData);
+        EventManager.getInstance().post(new SaveRequestEvent(
+                PrefsManager.getInstance().getSaveLocation(), e.personData, e.groupData));
     }
```
damithc: These two methods are counterparts. So method signatures should be similar, name format wise. e.g. 
`handleLocalModelChangedEvent`
`handleRemoteModelChangedEvent`

src/main/java/address/prefs/PrefsManager.java
```diff
@@ -0,0 +1,94 @@
...
+import java.io.File;
+
+/**
+ * Manages saving/retrieving of preferences on the user level.
+ * Publicly accessible singleton class.
+ */
+public class PrefsManager {
+
+    public static final String PREF_NODE_NAME_SAVE_LOC = "save-location";
+    public static final String PREF_NODE_NAME_MIRROR_LOC = "mirror-location";
```
HansNewbie: Maybe we don't need to mention `NODE_NAME` as the base implementation might change. `SAVE_LOC_PREF_KEY` might be better? I assume this is some key-value pairing data.

damithc:
No more comments from me

src/main/java/address/MainApp.java
damithc: Sure

HansNewbie:
No more comments.

src/main/java/address/storage/StorageManager.java
YijinL: `LocalModelSync` implies that the local model has changed and the cause is syncing with the remote model.

`RemoteModelChanged` only implies the remote model has changed (since our last check?). A similar event has already been implemented: `NewMirrorDataEvent`.

PR #92, opened 2016-05-18
Pending local changes #8

### LINKED ISSUE #8: Pending local changes
Changes made through the UI should be marked pending. Those changes can be reverted or finalised depending on new data from `SyncManager`.

m133225:
Fixes #8

`Person`, `ContactGroup` and `AddressBook` now has its `Model` version which implements a boolean status variable `isPending`. Unless necessary, components such as `EditDialog` should not have to deal with the `Model` version of it, and thus the data should be cast to its non-`Model` version before being passed to such components. Examples of components which may have to deal with the `Model` version may include the `PersonOverview` since we wish to have real-time updates of the pending status (to reflect if the add/edit/delete is successful).

This does not fully implement the desired changes since all data (except the AddressBook wrapper) is passed to the cloud for processing. <s>Moreover, for</s> now, <s>only</s> the list of `Person` **and list of `Group` are**<s>is</s> sent in the success 'notification'. This behaviour will thus affect all data here i.e. every `ModelPerson` data is made to !isPending even when only 1 of it is modified, which should not be the case.

When using GitHub's API, usually only the affected data is embedded within the request.

### DIFF
--- a/src/main/java/address/controller/GroupListController.java
+++ b/src/main/java/address/controller/GroupListController.java
@@ -1,6 +1,7 @@
 package address.controller;
 
 import address.model.ContactGroup;
+import address.model.ModelContactGroup;
 import address.model.ModelManager;
 import address.ui.GroupListViewCell;
 import javafx.collections.ObservableList;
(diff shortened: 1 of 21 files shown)


damithc:
@m133225 Let's discuss tomorrow before we go any further with this approach.

PR #102, opened 2016-05-19
Instantaneous loading of contact github user page by using tabs(hidden) #85

### LINKED ISSUE #85: Instantaneous loading of contact github user page by using tabs(hidden)
So that user could navigate faster without waiting for page to load every time.

Tabs should include the current selected contact and following two contacts in the list.

yl-coder:
Fixes #85


src/main/java/address/Browser/BrowserManager.java
```diff
@@ -1,53 +1,177 @@
...
     }
 
     /**
-     * Frees resources allocated to the browser.
+     * Initialise browser to initial state.
      */
-    public void freeBrowserResources(){
-        browser.dispose();
+    public void initBrowser() {
+        for (int i = 0; i < browser.getNumberOfPages(); i++){
```
HansNewbie: Maybe can load on just main page instead of on all three tabs since user may not check the other tabs.

src/main/java/address/Browser/BrowserManager.java
```diff
@@ -1,53 +1,177 @@
...
+                        System.out.println("finish loading frame");
+                    }
+                }
+            });
+        }
     }
 
     /**
-     * Frees resources allocated to the browser.
+     * Initialise browser to initial state.
```
HansNewbie: Initalises :p

src/main/java/address/Browser/BrowserManager.java
```diff
@@ -1,53 +1,177 @@
...
+     * Loads the person's GitHub profile page to the browser.
+     * PreCondition: ModelManager getFilteredPersons method returns more than one person.
+     * @param person
+     */
+    public void loadGitHubProfilePage(Person person) {
+
+        List<Person> filteredPersons = modelManager.getFilteredPersons();
+        int indexOfContactsInFilteredListView = filteredPersons.size();
+
+        int indexOfSelectedTabBrowserPage = findIndexOfTabBrowserThatContainsPersonGitHubWebPage(person);
```
m133225: This should be `Optional` instead of having to compare it with a magic number if it doesn't exist.

HansNewbie: `indexOfSelectedBrowserTab` or `indexOfSelectedTab`

HansNewbie: `findTabIndexOfBrowser...`

src/main/java/address/Browser/BrowserManager.java
```diff
@@ -1,53 +1,177 @@
...
+
+import java.util.List;
 
 /**
  * Manages the browser.
  */
 public class BrowserManager {
 
-    private Browser browser;
+    private static final int PERSON_DOES_NOT_EXISTS_IN_BROWSER = -1;
```
HansNewbie: `PERSON_INEXISTENT_IN_BROWSER`.

src/main/java/address/Browser/BrowserManager.java
yl-coder: noted. Changed

src/main/java/address/Browser/BrowserManager.java
```diff
@@ -1,53 +1,177 @@
...
 
-    private void automateClickingAndScrolling() {
-        DOMElement container = browser.getDocument().findElement(By.id("js-pjax-container"));
-        DOMElement link = browser.getDocument().findElement(By.className("octicon octicon-repo"));
-        if (link != null) {
-            container.addEventListener(DOMEventType.OnLoad, e ->
+    private int findASuitableBrowserTabAndLoadThePersonGitHubProfilePage(Person person, List<Person> persons,
+                                                                         int indexOfContactsInFilteredListView) {
+
+        int indexOfSelectedTabBrowserPage = 0;
```
m133225: I think here to line 114 can be extracted into another method e.g. `getUnassignedTabIndex`. Also try to simplify the logic if possible, since it seems a bit messy.

src/main/java/address/Browser/BrowserManager.java
yl-coder: ok changed. :p

src/main/java/address/Browser/BrowserManager.java
```diff
@@ -1,53 +1,177 @@
...
+     */
+    private int findIndexOfTabBrowserThatContainsPersonGitHubWebPage(Person person) {
+        int indexOfSelectedTabBrowserPage = PERSON_DOES_NOT_EXISTS_IN_BROWSER;
+        for (TabBrowserPage page : browser.getAllPages()) {
+            if (!page.isAssignedToLoadAPersonGitHubProfilePage()) {
+                continue;
+            }
+            if (page.getPerson().equals(person)) {
+                indexOfSelectedTabBrowserPage = browser.indexOf(page);
+                break;
```
HansNewbie: No need to `break` here. Just `return browser.indexOf(page)`. Then, just `return PERSON_DOES_NOT_EXISTS_IN_BROWSER` at the end of the method.

src/main/java/address/Browser/IdentifiableBrowser.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public Person getPerson(){
+        return this.person;
+    }
+
+    /**
+     * Checks if browser is assigned to load a person's GitHub profile page
+     * @return
+     */
+    public boolean isAssignedToLoadAPersonGitHubProfilePage(){
```
m133225: Maybe just name it `isAssigned`?

src/main/java/address/Browser/BrowserManager.java
```diff
@@ -1,53 +1,177 @@
...
+                || !nextTabBrowserPage.getPerson().equals(nextPersonInFilteredList)) {
+                nextTabBrowserPage.loadURL(nextPersonInFilteredList.getPersonGitHubProfilePageUrl(),
+                                           nextPersonInFilteredList);
+            }
+            indexOfNextTabBrowserPage = (indexOfNextTabBrowserPage + 1) % browser.getNumberOfPages();
+            indexOfNextPersonInFilteredList = (indexOfNextPersonInFilteredList + 1) % persons.size();
+        } while (indexOfNextTabBrowserPage != indexOfSelectedTabBrowserPage);
+    }
+
+    private void automateClickingAndScrolling(Browser browser) {
```
YijinL: Good to add some comments providing some context (clicking what? scrolling where? when does it behave differently?)

src/main/java/address/controller/MainController.java
```diff
@@ -363,15 +361,13 @@ public void releaseResourcesForAppTermination(){
         browserManager.freeBrowserResources();
     }
 
-    public void loadGithubProfilePage(String githubUserName){
-        browserManager.getBrowser().loadURL("https://www.github.com/" + githubUserName);
+    public void loadGitHubProfilePage(Person person){
+        browserManager.loadGitHubProfilePage(person);
```
HansNewbie: Maybe we can pass all the persons needed to be loaded here instead of coupling `browserManager` with `modelManager`.

src/main/java/address/Browser/TabBrowser.java
```diff
@@ -0,0 +1,60 @@
+package address.browser;
+
+import com.teamdev.jxbrowser.chromium.javafx.BrowserView;
+import javafx.scene.control.Tab;
+import javafx.scene.control.TabPane;
+
+import java.util.ArrayList;
+
+/**
+ * A TabPane UI that consists of a browser with multiple tabs.
```
HansNewbie: Shouldn't put implementation detail (`TabPane UI`) in comment because the implementation can be changed although the class still does the same thing. `A browser with multiple tabs` suffices.

src/main/java/address/model/Person.java
```diff
@@ -240,4 +240,8 @@ public String toString() {
         return "Person: " + getFullName();
     }
 
+    public String getPersonGitHubProfilePageUrl(){
```
m133225: `getProfilePageUrl` will do. This method is already in `Person` and we only have 1 kind of URL now.

src/main/java/address/Browser/BrowserManager.java
```diff
@@ -1,53 +1,177 @@
...
+                indexOfContactsInFilteredListView = pagePersonIndex;
+                indexOfSelectedTabBrowserPage = browser.indexOf(page);
+            }
+        }
+        browser.getAllPages().get(indexOfSelectedTabBrowserPage).loadURL(person.getPersonGitHubProfilePageUrl(),
+                person);
+        return indexOfSelectedTabBrowserPage;
+    }
+
+    private void updateOtherBrowserTabs(List<Person> persons, int indexOfSelectedTabBrowserPage) {
```
YijinL: Comment on what kind of updates etc.

Use standard javadoc header comments, especially elaborate on the `@param`s and `@return`s to clarify the input and output
```java
/**
 * <description>
 * @param persons <purpose and constraints>
 * @param selectedTabIndex <purpose>
 */
```

src/main/java/address/Browser/BrowserManager.java
YijinL: Alternatives: `PERSON_NOT_IN_BROWSER`, `PERSON_NOT_LOADED` etc depending on context

src/main/java/address/Browser/BrowserManager.java
```diff
@@ -1,53 +1,177 @@
...
+        updateOtherBrowserTabs(filteredPersons, indexOfSelectedTabBrowserPage);
+    }
+
+    /**
+     * Returns the index of the tab in the TabBrowser that contains the person's GitHub profile page.
+     * @param person
+     * @return the index of the tab in the TabBrowser that contains the person's GitHub profile page
+     *         or -1 if the tabs of the TabBrowser do not contain the person's GitHub profile page.
+     */
+    private int findIndexOfTabBrowserThatContainsPersonGitHubWebPage(Person person) {
```
HansNewbie: `findTabIndexOfPersonWebPageInBrowser()` or something similar.

src/main/java/address/Browser/BrowserManager.java
```diff
@@ -1,53 +1,177 @@
...
+        }
+        return indexOfSelectedTabBrowserPage;
     }
 
-    private void automateClickingAndScrolling() {
-        DOMElement container = browser.getDocument().findElement(By.id("js-pjax-container"));
-        DOMElement link = browser.getDocument().findElement(By.className("octicon octicon-repo"));
-        if (link != null) {
-            container.addEventListener(DOMEventType.OnLoad, e ->
+    private int findASuitableBrowserTabAndLoadThePersonGitHubProfilePage(Person person, List<Person> persons,
```
HansNewbie: A header comment here would help; what is `person` and what are `persons`?

YijinL: When the method returns will the page be loaded? or does it just find the tab index? You should define what "suitable" means too.

src/main/java/address/Browser/BrowserManager.java
yl-coder: ok noted. Changed for all

src/main/java/address/Browser/BrowserManager.java
yl-coder: Using Optional. No longer applicable.

damithc:
I tried the branch. There doesn't seem to be any difference in page loading speed.

src/main/java/address/Browser/IdentifiableBrowser.java
yl-coder: Right, i think the header comment should explain it clear enough.

src/main/java/address/Browser/BrowserManager.java
yl-coder: ok noted changed.

src/main/java/address/Browser/BrowserManager.java
yl-coder: ok noted. Changed.

src/main/java/address/controller/MainController.java
yl-coder: Right. Changed. Passed in needed instance through the constructor.

src/main/java/address/Browser/TabBrowser.java
yl-coder: Ok noted. Changed

src/main/java/address/model/Person.java
yl-coder: Right, Changed

src/main/java/address/Browser/BrowserManager.java
yl-coder: Ok noted. added

yl-coder:
@damithc You're right, i just tried it. Not working as normal. Something went wrong when i made some changes yesterday.

yl-coder:
It's working now. Have made changes based on review comment. 

However I may make some major changes on the abstraction. I should create another class called AddressBookBrowser or BrowserComponent. As the browser is designed specifically to load AddressBook contents faster, stuff like TabBrowser should be hidden away under that.

And also optimize on the algorithm of replacing pages by using Set.

src/main/java/address/Browser/BrowserManager.java
```diff
@@ -1,53 +1,193 @@
...
+                        System.out.println("finish loading frame");
+                    }
+                }
+            });
+        }
     }
 
     /**
-     * Frees resources allocated to the browser.
+     * Initialises browser to initial state.
```
HansNewbie: I think the header comment can be better. Maybe:
```
Initialises browser (no need to initial state as it is implied)

This includes:
- Loads instruction page on browser
```

We can add the list if there is anything else to initialise, in which then we abstract the load instruction page into another method.

YijinL:
You cannot retrieve elements from a `Set` (only can check `contains`), a workaround is using a `Map<E, E>`

HansNewbie:
http://stackoverflow.com/questions/12756360/how-to-make-java-set for set operations (union, intersection). We can stream the subset (say intersection) and operate on it. We can always use other data structure if `Set` is not suitable.

YijinL:
Also remember that `retainAll` and `removeAll` will mutate the set, so don't call it on the master collection.

src/main/java/address/browser/AddressBookBrowser.java
```diff
@@ -0,0 +1,273 @@
...
+     * Gets a list of person that are needed to be loaded to the browser.
+     * @param filteredPersons
+     * @param indexOfPerson
+     * @return
+     */
+    private ArrayList<Person> getListOfPersonsToBeLoaded(List<Person> filteredPersons, int indexOfPerson) {
+        ArrayList<Person> listOfPersonsToBeLoaded = new ArrayList<>();
+
+        for (int i = 0; i < BrowserManager.NUMBER_OF_PRELOADED_PAGE && i < filteredPersons.size(); i++){
+            listOfPersonsToBeLoaded.add(filteredPersons.get((indexOfPerson + i) % filteredPersons.size()));
```
HansNewbie: Does our list of person loop back to the top (i.e. when you reach bottom, press down, it goes to the person at the top)? If not, we are caching unnecessarily.

yl-coder: No, it doesn't. Should we implement that? or set a boundary for caching?

src/main/java/address/browser/AddressBookBrowser.java
```diff
@@ -0,0 +1,273 @@
...
+        }
+        return listOfPersonsToBeLoaded;
+    }
+
+    private List<Person> getFilteredPersons() {
+        return filteredModelPersons.stream()
+                .map(person -> (Person) person)
+                .collect(Collectors.toList());
```
HansNewbie: What's the use of this method? Seems that we can use `filteredModelPersons` directly since `ObservableList` inherits `List`.

src/main/java/address/browser/AddressBookBrowser.java
HansNewbie: Boundary is better. What we can do is maybe load the persons before the index if we reach boundary, e.g. we have 3 tabs and persons from index 0 to 10, if we are at index 9, we load 8, 9 and 10 since after the user reach index 10, he can only go back to 9 then 8.

YijinL: I think you intended the +2 system to be a placeholder till we work out a proper caching policy? This PR is for us to test caching in general?

HansNewbie: If that's the case, then sure.

src/main/java/address/browser/AddressBookBrowser.java
yl-coder: mapping of ModelPerson to Person. The live copy is kept in ModelPerson format.

YijinL: I think he is using this to downcast `ModelPerson` to `Person` probably because their equals methods are incompatible.
Btw YL if this is the case you need to `.map(Person::new)` because each object's true class and method behaviours still follow `ModelPerson`.

I will change them `ModelPerson` to a container class instead of a subclass of `Person` in another PR.

HansNewbie: Alright. The lambda variable name `person` confuses me.

src/main/java/address/util/Range.java
```diff
@@ -0,0 +1,29 @@
...
+     * @param val
+     * @return
+     */
+    public boolean isWithinRange(int val){
+
+        if (end > start) {
+            return val <= end && val >= start;
+        } else {
+            return val <= end || val >= start;
+        }
```
HansNewbie: What if `end == start`? In this implementation, it will cover everything. Should the range just be 1 number?

src/main/java/address/browser/AddressBookBrowser.java
yl-coder: Yes, we should wait till proper caching policy is defined.

src/main/java/address/browser/BrowserManager.java
```diff
@@ -1,53 +1,50 @@
...
+    public BrowserManager(ObservableList<ModelPerson> filteredModelPersons) {
+        this.filteredModelPersons = filteredModelPersons;
+        browser = new AddressBookBrowser(NUMBER_OF_PRELOADED_PAGE, this.filteredModelPersons);
+        browser.registerListeners();
     }
 
     /**
-     * Frees resources allocated to the browser.
+     * Loads the person's profile page to the browser.
+     * PreCondition: filteredModelPersons more than one person.
```
HansNewbie: `more than one person` or `one or more person`? The `browser` comment is `>= 1`, which is `one or more`.

src/main/java/address/browser/BrowserManager.java
```diff
@@ -1,53 +1,50 @@
...
-            public void onFinishLoadingFrame(FinishLoadingEvent finishLoadingEvent) {
-                automateClickingAndScrolling();
-            }
-        });
+    private ObservableList<ModelPerson> filteredModelPersons;
+
+    public BrowserManager(ObservableList<ModelPerson> filteredModelPersons) {
+        this.filteredModelPersons = filteredModelPersons;
+        browser = new AddressBookBrowser(NUMBER_OF_PRELOADED_PAGE, this.filteredModelPersons);
+        browser.registerListeners();
```
HansNewbie: Is it possible to move this to constructor of `AddressBookBrowser`?

src/main/java/address/browser/AddressBookBrowser.java
```diff
@@ -0,0 +1,273 @@
...
+
+/**
+ * A browser for displaying AddressBook contacts' profile page.
+ */
+public class AddressBookBrowser{
+
+    private static final String INSTRUCTION_PAGE_HTML_CODE =
+            "<html><body><h3>To view contact's web page, click on the contact on the left.</h3></body></html>";
+
+    private ArrayList<BrowserTab> pages;
```
HansNewbie: Maybe `browserTabs` is a better variable name?

src/main/java/address/browser/AddressBookBrowser.java
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    /**
+     * Initialises browser
+     *
+     * This includes:
+     - Loads instruction page on browser
+     */
+    public void init() {
+        pages.get(0).loadHTML(INSTRUCTION_PAGE_HTML_CODE);
```
HansNewbie: If we are not initialising everything here (`hideTabs` sounds like part of `init`, but it's separated), then we might not need `init()`. This method can be renamed to `loadInstructionToTab` or something similar.

src/main/java/address/browser/AddressBookBrowser.java
```diff
@@ -0,0 +1,273 @@
...
+    public TabPane getAddressBookBrowserView(){
+        return addressBookBrowserView;
+    }
+
+    /**
+     * Dispose the resources used by this browser. Once called, the browser is no longer usable.
+     */
+    public void dispose(){
+        for (int i = 0; i < pages.size(); i++) {
+            pages.get(i).dispose();
```
HansNewbie: Can we `stream`?

src/main/java/address/browser/AddressBookBrowser.java
yl-coder: @YijinL changing to .map(Person::new) causes .equals method to not work properly. I'm not too sure why though.

src/main/java/address/browser/AddressBookBrowser.java
```diff
@@ -0,0 +1,273 @@
...
+    private void loadProfilePages(ArrayList<Person> listOfPersonsToBeLoaded) {
+        for (int c = 0; c < listOfPersonsToBeLoaded.size(); c++){
+            Optional<BrowserTab> notAssignedTab = pages.stream().filter(page -> !page.isAssigned()).findFirst();
+            int indexOfTab = 0;
+
+            if(notAssignedTab.isPresent()){
+                indexOfTab = pages.indexOf(notAssignedTab.get());
+            }
+
+            pages.get(indexOfTab).loadProfilePage(listOfPersonsToBeLoaded.get(c));
```
HansNewbie: instead of using `indexOfTab`, we can use `Optional.orElse` from `notAssignedTab` here.

src/main/java/address/util/Range.java
yl-coder: Thats right. Noted and changed.

src/main/java/address/browser/AddressBookBrowser.java
HansNewbie: Wait, let me see again. I missed something.

src/main/java/address/browser/BrowserManager.java
yl-coder: Should be >= 1. That's a mistake. I have changed it to follow the same convention as browser

src/main/java/address/browser/BrowserManager.java
yl-coder: registerListeners could be optional? BrowserManager job is to register or not to register it?

src/main/java/address/browser/AddressBookBrowser.java
HansNewbie: Yes, you can check out http://www.oracle.com/technetwork/articles/java/java8-optional-2175753.html for other uses of `Optional`, but we can use `Optional.ifPresent` and `Optional.orElse` here.

src/main/java/address/browser/BrowserManager.java
HansNewbie: Sure.

src/main/java/address/browser/AddressBookBrowser.java
yl-coder: yup. right

src/main/java/address/browser/AddressBookBrowser.java
yl-coder: Right, renamed to loadInstruction.

src/main/java/address/browser/AddressBookBrowser.java
yl-coder: right, yes we can. Changed

src/main/java/address/browser/AddressBookBrowser.java
YijinL: @yl-coder I wasn't clear, uhh basically you cannot `.equals` a Person with a ModelPerson at all. 

If you are comparing `ModelPerson -- ModelPerson` then you should not convert with `new`, only if you are comparing `Person -- ModelPerson` then you should convert `ModelPerson` to `Person` with the `new Person(Person p)` constructor.

src/main/java/address/browser/AddressBookBrowser.java
```diff
@@ -0,0 +1,273 @@
...
+    /**
+     * Computes the range of index of the displayed AddressBook contact that needs to be added to the browser cache.
+     * @param listOfContactsDisplayed The list of contacts currently displayed.
+     * @param indexOfPersonInListOfContacts The index of the currently selected person in the AddressBook contact list.
+     * @return
+     */
+    private Range getRangeOfIndexToLoad(List<Person> listOfContactsDisplayed, int indexOfPersonInListOfContacts) {
+        Range rangeOfIndexToLoad;
+        if(BrowserManager.NUMBER_OF_PRELOADED_PAGE >= listOfContactsDisplayed.size()){
+            rangeOfIndexToLoad = new Range(indexOfPersonInListOfContacts, indexOfPersonInListOfContacts -1);
```
HansNewbie: This is a bit hard to understand if someone does not know how `Range` works. Maybe we can use `Range(0, listOfContactsDisplayed.size() - 1)` instead for readability?

src/main/java/address/browser/AddressBookBrowser.java
```diff
@@ -0,0 +1,273 @@
...
+        for (BrowserTab browserTab: pages) {
+            int personIndexInFilteredPerson = listOfContactsDisplayed.indexOf(browserTab.getPerson());
+
+            if (personIndexInFilteredPerson == -1) {
+                continue;
+            }
+
+            if (!rangeOfIndexToLoad.isWithinRange(personIndexInFilteredPerson)) {
+                browserTab.unloadProfilePage();
+            }
```
HansNewbie: Instead of using range which is hard to understand to determine who we would like to load, why don't this method just pass the list of `Person` that we would like to load to this method, and then unload any tabs which `Person` is not inside the list?

We can use `stream.map` for that, too.

HansNewbie: Then, we can stream all `Person` in our tabs as `listOfPersonAlreadyLoaded` since the rest has been unloaded and became `null` (beware of that). Then we won't need `Range`.

src/main/java/address/browser/AddressBookBrowser.java
HansNewbie: Btw, if there is no available tab, this will keep overriding the first tab in loading the profile page.

If that won't be the case due to the assumption that `no. of available tabs` `==` `size of listOfPersonsToBeLoaded`, we can stream and collect the tabs that's not assigned, then use [IntStream](http://stackoverflow.com/questions/22793006/java-8-foreach-with-index) to stream using indices.

yl-coder: I collected the browserTab reference instead of using intStream. But i guess it is the same thing since if I use intStream, i need to get the browserTab later also.

HansNewbie: Yep. I meant to collect the browserTab first then use intStream on loading the pages (because we would like 1-to-1 matching loading between a Person and a tab).

src/main/java/address/browser/AddressBookBrowser.java
yl-coder: new Range(indexOfPersonInListOfContacts, indexOfPersonInListOfContacts -1); 
and 
new Range(0, listOfContactsDisplayed.size() - 1)

are two different values?

HansNewbie: Yes, but don't they cover the same range? Since we are doing inclusive check.

src/main/java/address/browser/AddressBookBrowser.java
yl-coder: You're right, I'm comparing it with ModelPerson -- Person. That's why it was not working. It works now. Thanks

yl-coder:
Thanks all for the review

PR #103, opened 2016-05-19
Fix testing & CI #101

### LINKED ISSUE #101: Fix testing & CI
- `headless` appears to always run without failures on Travis
  - `GroupEditDialogTest` should be failing at the moment, but there is no error on Travis
- headful tests seem to get stuck on Travis (possibly because of forking?)
- running headless guitests seems to cause problem in JVM (at least on my local machine)

m133225:
Fixes #101

Ongoing.


m133225:
The tests should be working now.
1) A few workarounds have been implemented:
- For GUI Tests, we **require a dummy `$TEMP_ADDRESS_BOOK`** for the application to start successfully. 
  - We are **unable to click on the `Alert` dialog** since it launches and blocks the application (during `FxToolKit.setupApplication(TestApp.class)`) before its `Stage` is passed to the robot for automated control.
  - We are **unable to specify a custom data file for testing** as of yet, since `Config`(where we define custom test settings) is not the component handling the data file path, but `Prefs`. Moving this path information from `Prefs` to `Config` can be implemented soon in the near future, so that we do not need the dummy file.

2) In addition, interactions between **JxBrowser and headless tests on Mac OSX** causes the JVM to **crash** with a SIGSEGV. However, headful tests should still be working fine.

3) Checkstyle and headless tests run on Travis at the moment.
**Even if checkstyle fails, the headless tests will still be run.** However, the success of the test may be difficult/troublesome to determine as we have to sieve through the log for a `:headless` that does not fail.

It might be better to remove checking of style on Travis since the status of the tests will be reflected immediately. What do you guys think?

damithc:
Good progress Martin. Let's continue until we are totally on top of the situation.
I'm ok to suspending checkstyle and PMD for now.
Yes, we should improve the Config object to be able to control more things. We can also move prefs from registry to a normal file so that it can be changed to the values we want before running the tests.
At some point, we may be able to get the help from jxbrowser guys if the solution eludes us.

damithc:
My results: headless kept failing a few times, but after a while both modes started to pass consistently. Not sure why it failed initially.

damithc:
There's something funny going on. Sometimes `gradlew headful` doesn't even open the GUI. Just says pass. But sometimes it does open the GUI.

m133225:
I think that's because you have already ran the `headful` tests before using the same build, so gradle skips running it.
`./gradlew clean headful` should allow the project to be re-built and re-tested (and thus shows the GUI).

damithc:
You are right. Adding `clean` did the trick.

m133225:
Maybe `clean` should automatically be run immediately after the tests in `headful` and `headless`? Just like HT's recommended `./gradlew check clean`, but we simply would need `./gradlew headful` to perform the same operations.

damithc:
headless tests timeout occasionally, usually the first time I run them after a while. They may be a timing issue e.g. click something before it appears.

m133225:
Yes, same for my Windows desktop. I think it is due to `TestFX`'s `FxToolkit.setupApplication` slow initialization. Will look into it.

m133225:
@damithc Could this be merged soon? There are already alot of changes (especially for the browser components which will often cause merge conflicts). Will work on mocking/stubbing and introducing more tests in another PR.

damithc:
Sure. Go ahead.

damithc:
After merging, may be get Thien to check if it works in his Linux comp?

PR #106, opened 2016-05-20
Create a release JAR #90

### LINKED ISSUE #90: Create a release JAR
Create a release JAR which contains all JARs inside, but will explode them on first run and the libraries can be updated individually. It is sort of a wrapper.

The release JAR also downloads the relevant jxBrowser before launching the main application. It acts as a wrapper.

The downside is that it takes more disk space in the sense that the wrapper has all JARs (meaning will be quite big, say 5MB for now). However, this wrapper can be replaced by main application JAR in the following update 😬

HansNewbie:
Fixes #90

### DIFF
--- a/docs/Release.md
+++ b/docs/Release.md
@@ -1,0 +1,37 @@
+#Release
+In adding updating function to addressbook, its release is tweaked to separate the dependencies libraries (their JARS)
+from the main app, so that only affected JARs need to be updated. This saves bandwidth in downloading update.
+
+At the same time, to give convenience to user in getting the application, the whole JARs of main app and the libraries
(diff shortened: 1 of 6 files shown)


docs/Release.md
```diff
@@ -0,0 +1,38 @@
+#Release
+In adding updating function to addressbook, its release is tweaked to separate the dependencies libraries (their JARS)
+from the main app, so that only affected JARs need to be updated. This saves bandwidth in downloading update.
```
damithc: Explain how things are now (and why), rather than how you arrived there. This phrase sound like you are describing how you arrived there `In adding updating function to addressbook, its release is tweaked`

PR #127, opened 2016-05-26
Make the cloud's API similar to GitHub's #94

### LINKED ISSUE #94: Make the cloud's API similar to GitHub's
We now send all data to the cloud for processing, which performance gets worse as the addressbook becomes bigger. Also, this is unlike the GitHub's API which we are trying to emulate for HT.

This will also bring in the desired purpose of #8 - which includes showing pending elements statuses

m133225:
Fixes #94.

Ongoing.

- [X] Implement `CloudService`, responsible for high-level API
- [X] Implement `CloudSimulator`, responsible for providing similar cloud API as GitHub
- [X] Handle conversion of cloud's response to local response
- [X] Rename `ContactGroup` to `Tag`
- [X] Add random chance of simulating loss of data/bad response code
- [X] Read/write from mirror file
- [X] Read latest cloud data using the new API (only if the appropriate mirror file has been created)
- [X] Implement API quota counter & its hourly reset
- [X] Check for bad response code from Cloud's response
- [X] Add random delay to `CloudSimulator`'s response
- [X] Differentiate cloud's and local model's `RateLimitStatus`
- [x] Handle multiple address books (including reading/writing to mirror file)
- [x] Add random chance of simulating data mutation/addition
- [x] Implement ETag
- [x] Add paged-requests
- [x] Update with documentations/comments & Refactoring

- Writes updates to cloud using the new API  
This should be handled by the model implementation. Will be temporarily disabled when this PR is merged.


m133225:
Should the `CloudSimulator` be separated into `CloudService`, which deals with the communication aspect, and `Cloud` (I guess this could still be called `CloudSimulator`), which verifies the validity of data & requests? Currently it is doing both converting into a `CloudResponse` as well as verifying of the data integrity, which may cause the class to be bloated.

i.e.
> ... -> SyncManager <---> CloudService <---> Cloud (Simulator)

damithc:
Yes, have as many classes as you need within that component.

src/main/java/address/sync/CloudSimulator.java
```diff
@@ -1,129 +1,670 @@
...
+            e.printStackTrace();
+            return null;
+        }
+    }
+
+    private int getNumberOfRequestsRequired(int dataSize, int resourcesPerPage) {
+        return (int) Math.ceil((double) dataSize / resourcesPerPage);
+    }
+
+    private void deletePersonFromData(List<CloudPerson> personList, String firstName, String lastName)
```
m133225: This does not delete the person from the list directly. It simply marks the person as `deleted`. However, the local model does not account for this (yet) & the `CloudService` simply ignores the deleted field.

Note that removing the person from the list directly may also cause problems, such as not being able to determine what has been deleted on the cloud without downloading the full list of data.

src/main/java/address/sync/CloudService.java
```diff
@@ -0,0 +1,394 @@
...
+        if (!isValid(cloudResponse)) {
+            return getResponseWithNoData(cloudResponse, headerHashMap);
+        }
+        return new ExtractedCloudResponse<>(cloudResponse.getResponseCode(), getRateLimitFromHeader(headerHashMap),
+                                            getRateRemainingFromHeader(headerHashMap),
+                                            getRateResetFromHeader(headerHashMap), null);
+    }
+
+    /**
+     * Gets the list of persons for addressBookName, which have been modified after a certain time
```
m133225: `if quota is available`

src/main/java/address/sync/CloudService.java
```diff
@@ -0,0 +1,394 @@
...
+public class CloudService implements ICloudService {
+    private static final int RESOURCES_PER_PAGE = 100;
+
+    private final CloudSimulator cloud;
+
+    public CloudService(boolean shouldSimulateUnreliableNetwork) {
+        cloud = new CloudSimulator(shouldSimulateUnreliableNetwork);
+    }
+
+    public static void main(String[] args) {
```
m133225: Should be removed.

src/main/java/address/sync/CloudSimulator.java
```diff
@@ -1,129 +1,670 @@
...
 
-    private boolean simulateUnreliableNetwork = false;
-    private static final Random RANDOM_GENERATOR = new Random();
+    private RateLimitStatus rateLimitStatus;
+    private boolean shouldSimulateUnreliableNetwork;
+    private TickingTimer timer;
 
-    public CloudSimulator(boolean shouldSimulateUnreliableNetwork) {
-        this.simulateUnreliableNetwork = shouldSimulateUnreliableNetwork;
+    CloudSimulator(boolean shouldSimulateUnreliableNetwork) {
```
HansNewbie: Is the removal of the access modifier intended?

m133225: Yes, I think the whole of the cloud should only be exposed to members in the `sync` package. I find either way acceptable though.

src/main/java/address/sync/CloudService.java
```diff
@@ -0,0 +1,394 @@
...
+    @Override
+    public ExtractedCloudResponse<Void> createAddressBook(String addressBookName) throws IOException {
+        RawCloudResponse cloudResponse = cloud.initializeAddressBook(addressBookName);
+        HashMap<String, Long> headerHashMap = getHashMapFromHeader(cloudResponse.getHeaders());
+        if (!isValid(cloudResponse)) {
+            return getResponseWithNoData(cloudResponse, headerHashMap);
+        }
+        return new ExtractedCloudResponse<>(cloudResponse.getResponseCode(), getRateLimitFromHeader(headerHashMap),
+                                            getRateRemainingFromHeader(headerHashMap),
+                                            getRateResetFromHeader(headerHashMap), null);
```
HansNewbie: It seems the return value here is the same as `getResponseWithNoData`. Can they be mixed? like, just `return getResponseWithNoData()` instead of the `if`.

src/main/java/address/sync/CloudSimulator.java
```diff
@@ -1,129 +1,670 @@
...
+    }
+
+    private void modifyCloudTagListBasedOnChance(List<CloudTag> cloudTagList) {
+        cloudTagList.stream()
+                .forEach(cloudTag -> modifyCloudTagBasedOnChance(cloudTag));
+    }
+
+    private void addCloudPersonsBasedOnChance(List<CloudPerson> personList) {
+        for (int i = 0; i < MAX_NUM_PERSONS_TO_ADD; i++) {
+            if (RANDOM_GENERATOR.nextDouble() <= ADD_PERSON_PROBABILITY) {
```
m133225: This should also check whether `shouldSimulateUnreliableNetwork` is true. Currently, it ALWAYS has a chance to mutate.

src/main/java/address/sync/CloudService.java
```diff
@@ -0,0 +1,394 @@
...
+    private int getRateLimitFromHeader(HashMap<String, Long> header) {
+        return header.get("X-RateLimit-Limit").intValue();
+    }
+
+    private int getRateRemainingFromHeader(HashMap<String, Long> header) {
+        return header.get("X-RateLimit-Remaining").intValue();
+    }
+
+    private long getRateResetFromHeader(HashMap<String, Long> header) {
+        return header.get("X-RateLimit-Limit");
```
HansNewbie: Is this correct key? Looks the same with `getRateLimitFromHeader` except not converted to `intValue()`. Maybe if that's the case, `getRateLimitFromHeader` should call this method in which it will call `intValue()`.

src/main/java/address/sync/CloudService.java
```diff
@@ -0,0 +1,394 @@
...
+
+        return convertedList;
+    }
+
+    private Person convertToPerson(CloudPerson cloudPerson) {
+        Person person = new Person(cloudPerson.getFirstName(), cloudPerson.getLastName());
+        person.setStreet(cloudPerson.getStreet());
+        person.setCity(cloudPerson.getCity());
+        person.setPostalCode(cloudPerson.getPostalCode());
+        return person;
```
HansNewbie: How about birthday field?

src/main/java/address/sync/CloudService.java
m133225: Yep, this is a mistake. Should instead be `X-RateLimit-Reset`.

src/main/java/address/sync/CloudService.java
m133225: Yep that's missing. Also `githubUserName` and `tags`. 😢

src/main/java/address/sync/CloudSimulator.java
```diff
@@ -1,129 +1,670 @@
...
+     * Consumes 1 + floor(persons size/resourcesPerPage) API usage
+     *
+     * @param addressBookName
+     * @param resourcesPerPage
+     * @return
+     */
+    @Override
+    public RawCloudResponse getPersons(String addressBookName, int resourcesPerPage) {
+        if (shouldSimulateNetworkFailure()) return getNetworkFailedResponse();
+        if (shouldSimulateSlowResponse()) delayRandomAmount();
```
HansNewbie: Maybe we can abstract this out as `simulateSlowResponseRandomly()`.

HansNewbie: Not that important though.

src/main/java/address/sync/CloudService.java
```diff
@@ -0,0 +1,394 @@
...
+        if (!isValid(cloudResponse)) {
+            return getResponseWithNoData(cloudResponse, headerHashMap);
+        }
+        return new ExtractedCloudResponse<>(cloudResponse.getResponseCode(), getRateLimitFromHeader(headerHashMap),
+                                            getRateRemainingFromHeader(headerHashMap),
+                                            getRateResetFromHeader(headerHashMap), null);
+    }
+
+    @Override
+    public ExtractedCloudResponse<Void> createAddressBook(String addressBookName) throws IOException {
```
m133225: This is an API not in GitHub. We require this since there is no way to add an addressbook without having to manually adding a cloud file. Should this also cost API usage?

src/main/java/address/sync/CloudSimulator.java
```diff
@@ -1,129 +1,670 @@
...
+            return getEmptyResponse(HttpURLConnection.HTTP_NO_CONTENT);
+        } catch (NoSuchElementException e) {
+            return getEmptyResponse(HttpURLConnection.HTTP_BAD_REQUEST);
+        } catch (JAXBException e) {
+            return getEmptyResponse(HttpURLConnection.HTTP_INTERNAL_ERROR);
+        }
+    }
+
+    @Override
+    public RawCloudResponse initializeAddressBook(String addressBookName) {
```
m133225: This is an API not in GitHub. We require this since there is no way to add an addressbook without having to manually add a cloud file. Should this also cost API usage?

src/main/java/address/sync/model/CloudPerson.java
```diff
@@ -0,0 +1,117 @@
...
+    private String lastName;
+    private String street;
+    private String city;
+    private String postalCode;
+    private List<CloudTag> tags;
+    private boolean isDeleted;
+    private LocalDateTime lastUpdatedAt;
+
+    public CloudPerson() {
+        setLastUpdatedAt(LocalDateTime.now());
```
m133225: This should be removed since the empty constructor should not be used aside from reading and writing into XML files. Same for `CloudTag`.

src/main/java/address/sync/task/CloudUpdateTask.java
```diff
@@ -28,13 +28,13 @@ public void run() {
...
-        allData.addAll(groupsData);
+        allData.addAll(tagsData);
 
         File mirrorFile = PrefsManager.getInstance().getMirrorLocation();
-        try {
-            simulator.requestChangesToCloud(mirrorFile, personsData, groupsData);
-        } catch (JAXBException e) {
-            System.out.println("Error requesting changes to the cloud");
-        }
+//        try {
```
m133225: Should add in a comment to explain why this is temporarily disabled:
`The way the local model should use the API to push its update is not yet finalized.`

src/test/java/address/guiunittests/PersonEditDialogUnitTest.java
```diff
@@ -57,40 +57,40 @@ public void start(Stage primaryStage) {
 
     @Before
     public void setup() {
-        groupList = find("#groupList");
-        groupSearch = find("#groupSearch");
-        groupResults = find("#groupResults");
+        tag = find("#tag");
```
m133225: `tagList`

src/main/java/address/sync/CloudSimulator.java
HansNewbie: Yes. I think GitHub has an API to open a repo. Maybe make it similar to that?

src/main/java/address/sync/CloudSimulator.java
```diff
@@ -1,129 +1,670 @@
...
+        ZonedDateTime zonedDateTime = localDateTime.atZone(ZoneOffset.systemDefault());
+        return zonedDateTime.getOffset();
+    }
+
+    private boolean shouldSimulateNetworkFailure() {
+        return shouldSimulateUnreliableNetwork && RANDOM_GENERATOR.nextDouble() <= FAILURE_PROBABILITY;
+    }
+
+    private boolean shouldSimulateSlowResponse() {
+        return shouldSimulateUnreliableNetwork && RANDOM_GENERATOR.nextDouble() <= FAILURE_PROBABILITY;
```
HansNewbie: Maybe have another variable called `NETWORK_CONGESTION_PROBABILITY` or something, to differentiate with network failure. Also, this can have higher chance than failure.

src/main/java/address/sync/CloudSimulator.java
```diff
@@ -1,129 +1,670 @@
...
+            timer.stop();
+        }
+        int timeout = (int) (rateLimitStatus.getQuotaReset() - LocalDateTime.now().toEpochSecond(getSystemTimezone()));
+        timer = new TickingTimer("Cloud Quota Reset Time", timeout, this::printTimeLeft,
+                                this::resetQuotaAndRestartTimer, TimeUnit.SECONDS);
+        timer.start();
+    }
+
+    private void printTimeLeft(int timeLeft) {
+        System.out.println(timeLeft + " seconds remaining to quota reset.");
```
m133225: Should be removed or print only occasionally e.g. every 60 seconds.

src/main/java/address/sync/CloudSimulator.java
```diff
@@ -1,129 +1,670 @@
...
+        System.out.println("Current quota left: " + rateLimitStatus.getQuotaRemaining() + ", using " + quotaUsed);
+        return quotaUsed <= rateLimitStatus.getQuotaRemaining();
+    }
+
+    private boolean isExistingPerson(List<CloudPerson> personList, CloudPerson targetPerson) {
+        return personList.stream()
+                .filter(person -> person.getFirstName().equals(targetPerson.getFirstName())
+                        && person.getLastName().equals(targetPerson.getLastName()))
+                .findAny()
+                .isPresent();
```
HansNewbie: I think it would be better if we use `personList.contains(targetPerson)` with `CloudPerson` having equality implemented. This would be helpful in the case, say, `CloudPerson` now is compared using their `generatedId`.

src/main/java/address/sync/CloudSimulator.java
```diff
@@ -1,129 +1,670 @@
...
+        String newPersonFirstName = newPerson.getFirstName();
+        String newPersonLastName = newPerson.getLastName();
+        if (newPersonFirstName == null || newPersonLastName == null) {
+            throw new IllegalArgumentException("Fields cannot be null");
+        }
+        if (isExistingPerson(personList, newPerson)) throw new IllegalArgumentException("Person already exists");
+
+        personList.add(newPerson);
+
+        return newPerson;
```
HansNewbie: I think this method is outdated. The comment says `return null if person not added` but I don't see it in the implementation. I also don't see anything that would throw `JAXBException` and instead it throws many `IllegalArgumentException`.

src/main/java/address/sync/CloudSimulator.java
m133225: Okay, found it in the GitHub API list 😢

src/main/java/address/sync/CloudSimulator.java
m133225: Hmm, this is not checking for complete equality but for a unique "key". Equality could be used in other places such as checking for differences between 2 objects (in other fields).

src/main/java/address/sync/CloudSimulator.java
```diff
@@ -1,129 +1,670 @@
...
+    private void deletePersonFromData(List<CloudPerson> personList, String firstName, String lastName)
+            throws NoSuchElementException {
+        CloudPerson deletedPerson = getPersonIfExists(personList, firstName, lastName);
+        deletedPerson.setDeleted(true);
+    }
+
+    private CloudTag addTag(List<CloudTag> tagList, CloudTag newTag) {
+        if (newTag == null) throw new IllegalArgumentException("Tag cannot be null");
+        String tagName = newTag.getName();
+        if (tagName == null) throw new IllegalArgumentException("Fields cannot be null");
```
HansNewbie: I think we can use `newTag.getName() == null` here since `tagName` not used anywhere else.

src/main/java/address/sync/CloudSimulator.java
HansNewbie: Noted. Can be kept.

m133225:
Requests for review before I merge this!

Note that the mirror data will not be functioning after this PR is merged, as `SyncManager` does not make use of the new API yet.

src/main/java/address/controller/TagEditDialogController.java
```diff
@@ -1,23 +1,23 @@
...
 import javafx.scene.control.TextField;
 
-public class GroupEditDialogController extends EditDialogController {
+public class TagEditDialogController extends EditDialogController {
     @FXML
-    private TextField groupNameField;
+    private TextField tagNameField;
 
-    private ContactGroup finalGroup;
+    private Tag finalTag;
```
yl-coder: I think editedTag may be a better name. I guess the method getFinalInput could be changed also?

src/main/java/address/util/JsonUtil.java
```diff
@@ -3,12 +3,15 @@
...
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.databind.type.TypeFactory;
 
 import java.io.IOException;
+import java.util.Collection;
 import java.util.HashMap;
+import java.util.HashSet;
```
HansNewbie: Only added import in this class, seems to be unused.

src/main/java/address/model/PersonEditDialogTagsModel.java
```diff
@@ -0,0 +1,136 @@
...
+public class PersonEditDialogTagsModel {
+    List<SelectableTag> tags = new ArrayList<>();
+    List<SelectableTag> filteredTags = new ArrayList<>();
+    Optional<Integer> selectedTagIndex = Optional.empty();
+
+    List<SelectableTag> assignedTags = new ArrayList<>();
+
+    public PersonEditDialogTagsModel(List<Tag> tags, List<Tag> assignedTags) {
+        List<SelectableTag> selectableTargets = tags.stream()
+                                                                .map(SelectableTag::new)
```
yl-coder: misalignment

src/main/java/address/sync/CloudService.java
```diff
@@ -0,0 +1,448 @@
...
+     * @throws IOException if content cannot be interpreted
+     */
+    @Override
+    public ExtractedCloudResponse<List<Person>> getPersons(String addressBookName) throws IOException {
+        int curPageNumber = 1;
+        RawCloudResponse cloudResponse;
+        List<CloudPerson> cloudPersons = new ArrayList<>();
+        do {
+            cloudResponse = cloud.getPersons(addressBookName, curPageNumber, RESOURCES_PER_PAGE, null);
+            cloudPersons.addAll(getDataListFromBody(cloudResponse.getBody(), CloudPerson.class));
```
HansNewbie: What happens if during the loop the cloud response above is broken or not valid?

src/main/java/address/sync/CloudSimulator.java
```diff
@@ -1,129 +1,750 @@
...
+            writeCloudAddressBookToFile(addressBookName, fileData);
 
-            // no data could be retrieved
-            if (RANDOM_GENERATOR.nextDouble() <= FAILURE_PROBABILITY) {
-                System.out.println("Cloud simulator: failure occurred! Could not retrieve data");
-                return Optional.empty();
-            }
+            modifyCloudPersonBasedOnChance(returnedPerson);
+            InputStream bodyStream = convertToInputStream(returnedPerson);
+            String ETag = RawCloudResponse.getETag(bodyStream, false);
```
HansNewbie: our naming convention would be `eTag` instead of `ETag`, unless there is something before `E`, like `previousETag`.

src/main/java/address/sync/CloudService.java
m133225: Haven't considered that case. The loop should end early with an empty response, I think.

Will resolve that in a new PR.

HansNewbie:
no more comment from me.

src/main/java/address/sync/CloudService.java
m133225: Issue raised in #163.

PR #129, opened 2016-05-26
Implement header status bar (actions) #110

### LINKED ISSUE #110: Implement header status bar (actions)
To add in a Status bar on the top of the UI to show what are the recent actions user had done.

yl-coder:
Fixes #110


HansNewbie:
I think having an event for each action just for UI is sort of wasteful. It might be better to fire such event when the model changes, e.g.:

`ContactDeletedEvent` is a subclass of `LocalModelChangedEvent` which is fired by `ModelManager`.

Hence, when it is fired, handler to `LocalModelChangedEvent` will be notified (say `StorageManager` will save the model) and handler to `ContactDeletedEvent` can handle such specific event (say `HeaderStatusBar` can show the action).

src/main/java/address/controller/StatusBarHeaderController.java
```diff
@@ -13,26 +12,33 @@
...
         footerStatusBar = new StatusBar();
     }
 
     public StatusBar getFooterStatusBarView() {
         return footerStatusBar;
     }
 
-    @Subscribe
-    public void handleContactEditedEvent(ContactEditedEvent e) {
+    public void postStatus(ContactBaseStatus status) {
```
YijinL: Would it be better to rename the below methods (`postContactEditedStatus`, `postContactDeletedStatus` etc) to `postStatus`? Basically overloading `postStatus` instead of needing this method to multiplex it.

src/main/java/address/controller/StatusBarHeaderController.java
```diff
@@ -13,26 +12,33 @@
...
+
+        if (status instanceof ContactCreatedStatus) {
+            postContactCreatedStatus((ContactCreatedStatus) status);
+        } else if (status instanceof ContactEditedStatus) {
+            postContactEditedStatus((ContactEditedStatus) status);
+        } else if (status instanceof ContactDeletedStatus) {
+            postContactDeletedStatus((ContactDeletedStatus) status);
+        }
+    }
```
YijinL: Would it be better to rename the below methods (`postContactEditedStatus`, `postContactDeletedStatus` etc) to `postStatus`? Basically overloading `postStatus` so we don't need this method to multiplex by type.

PR #134, opened 2016-05-27
Write test cases for browser component #112

### LINKED ISSUE #112: Write test cases for browser component
Write test cases for the browser component. GuiTest to test if the tabs load the correct web page and if the browser works for different number of tabs. Also write unit test to check if the page replacement algorithm works as intended.

yl-coder:
Fixes #112

### DIFF
--- a/src/test/java/address/unittests/AddressBookBrowserTest.java
+++ b/src/test/java/address/unittests/AddressBookBrowserTest.java
@@ -1,0 +1,104 @@
+package address.unittests;
+
+import address.browser.AddressBookBrowser;
+import address.model.Person;
+import address.util.JavafxThreadingRule;
+import address.util.TestUtil;
+import javafx.collections.FXCollections;
+import javafx.collections.ObservableList;
+
+import org.junit.Rule;
+import org.junit.Test;
+
+import static junit.framework.Assert.assertNull;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import java.util.List;
+
+/**
+ * Test cases to test AddressBookBrowser
+ */
+public class AddressBookBrowserTest {
+
+    @Rule
+    /**
+     * To run test cases on JavaFX thread.
+     */
+    public JavafxThreadingRule javafxRule = new JavafxThreadingRule();
+
+    @Test
+    public void testLoadProfilePage_oneTab_browserIsAssignedToLoadPerson(){
+        ObservableList<Person> list = FXCollections.observableArrayList();
+        list.setAll(TestUtil.generateSampleData());
+        AddressBookBrowser browser = new AddressBookBrowser(1, list);
+
+        browser.loadProfilePage(list.get(1));
+        assertTrue(browser.getBrowserTabs().get(0).getPerson().equals(list.get(1)));
+
+        browser.loadProfilePage(list.get(4));
+        assertTrue(browser.getBrowserTabs().get(0).getPerson().equals(list.get(4)));
+    }
+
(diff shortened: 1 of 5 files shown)


HansNewbie:
`JavaFxThreadingRule` should be in `test` folder. By our convention, the `x` should be in lower case. If `JavaFX` should be treated as one term, then should be `Javafx` instead of `JavaFx`. Not sure which one we should follow.

yl-coder:
Ok, created an util folder for it.

src/test/java/address/unittests/BrowserManagerTest.java
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Tests for the BrowserManager correct behaviours
+ */
+public class BrowserManagerTest {
+
+    @Rule
+    /**
+     * To run test cases on JavaFX thread.
+     */
+    public JavafxThreadingRule javaFxRule = new JavafxThreadingRule();
```
HansNewbie: `javafxRule`

src/test/java/address/unittests/AddressBookBrowserTest.java
```diff
@@ -0,0 +1,104 @@
...
+        List<Person> sampleList = TestUtil.generateSampleData();
+        ObservableList<Person> list = FXCollections.observableArrayList();
+        list.add(sampleList.get(0));
+        AddressBookBrowser browser = new AddressBookBrowser(3, list);
+
+        // Loads index 0 of list only.
+        browser.loadProfilePage(list.get(0));
+        assertTrue(list.get(0).equals(browser.getBrowserTabs().get(0).getPerson()));
+        assertFalse(list.get(0).equals(browser.getBrowserTabs().get(1).getPerson()));
+        assertFalse(list.get(0).equals(browser.getBrowserTabs().get(2).getPerson()));
```
HansNewbie: Shouldn't we assert the person in the tab as `null` here?

yl-coder: browser.getBrowserTabs().get(1).getPerson() == null?

src/test/java/address/unittests/BrowserManagerTest.java
```diff
@@ -0,0 +1,76 @@
...
+        browser.loadProfilePage(list.get(0));
+
+        //edit person in model
+        list.get(0).setGithubUserName("1");
+        browserManager.handleLocalModelChangedEvent(
+                new LocalModelChangedEvent(new ArrayList<Person>(), new ArrayList<ContactGroup>()));
+        //Check if person in the browser is edited.
+        assertTrue(browser.getBrowserTabs().stream().filter(browserTab
+                -> browserTab.getPerson() != null
+                   && browserTab.getPerson().getGithubUserName().equals("1")).findAny().isPresent());
```
HansNewbie: Should assert before changing that `list.get(0).getGithubUserName` is not `1`. If the sample data is ever changed such that the `githubUserName` is `1` even before changing, this test does not prove anything.

src/test/java/address/unittests/BrowserManagerTest.java
yl-coder: ok, fixed

src/test/java/address/unittests/AddressBookBrowserTest.java
HansNewbie: There is `assertNull` if you want to explore. What I am saying is if only 1 person should be loaded, shouldn't the other two be null? If not, then let's keep it this way.

Issue #121, opened 2016-05-25
Implement ViewableModel -- Model dual layer for Person objects


YijinL:
For allowing optimistic UI updating and separating session-level state from base data.

- The `ModelManager` will act as a facade and handle most events and method calls from other components
- The UI will observe the `Viewable` model for changes
- `Viewable` model objects will observe the backing model for changes
- Sync-downs will work on the backing model only, which then propogates up to the `Viewables`
- Saving and loading from disc will involve only the backing model

YijinL:
Tag objects not included in this proof of concept, will implement in future.

PR #136, opened 2016-05-27
Use ID for Person equality #135

### LINKED ISSUE #135: Use ID for Person equality
Also add person retrieving (as read-only Person superinterface) by ID capabilities to `ModelManager` so external components cannot change the in-model Person objects.

Model/Remote should control IDs of newly added Persons.

YijinL:
Resolves #135


YijinL:
should redo this after #121 is done

PR #148, opened 2016-05-31
Dependency check #147

### LINKED ISSUE #147: Dependency check
MainApp should do dependency check (that all required JARs are present) before running application.

HansNewbie:
Fixes #147

### DIFF
--- a/src/main/java/address/MainApp.java
+++ b/src/main/java/address/MainApp.java
@@ -13,9 +13,21 @@ import address.sync.SyncManager;
 import address.updater.UpdateManager;
 import address.util.Config;
 
+import address.util.FileUtil;
+import address.util.OsDetector;
 import javafx.application.Application;
 import javafx.application.Platform;
+import javafx.scene.control.Alert;
 import javafx.stage.Stage;
+
+import java.io.*;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.jar.Attributes;
+import java.util.jar.Manifest;
+import java.util.stream.Collectors;
 
 /**
  * The main entry point to the application.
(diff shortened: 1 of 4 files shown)


src/main/java/address/MainApp.java
```diff
@@ -49,13 +61,87 @@ public void init() throws Exception {
...
+            System.out.println("Manifest can't be read, not running dependency check");
+            e.printStackTrace();
+            return;
+        }
+
+        Attributes attr = manifest.getMainAttributes();
+        String classPath = attr.getValue("Class-path");
+        List<String> dependencies = new ArrayList<>(Arrays.asList(classPath.split("\\s+")));
+
+        processDependenciesException(dependencies);
```
HansNewbie: Will rename the method not to use `exception` as it is not java exception.

src/main/java/address/MainApp.java
```diff
@@ -49,13 +61,87 @@ public void init() throws Exception {
...
+
+        List<String> missingDependencies = dependencies.stream()
```
yl-coder: extract method?

src/main/java/address/MainApp.java
```diff
@@ -49,13 +61,87 @@ public void init() throws Exception {
...
+                .filter(dependency -> !FileUtil.isFileExists(dependency)).collect(Collectors.toList());
+
+        if (missingDependencies.isEmpty()) {
+            System.out.println("All dependencies are present");
+        } else {
+            String message = "Missing dependencies:\n";
+            for (String missingDependency : missingDependencies) {
+                message += "- " + missingDependency + "\n";
+            }
+            System.out.println(message.trim());
```
yl-coder: extract method for this block?

src/main/java/address/MainApp.java
HansNewbie: Ok.

src/main/java/address/MainApp.java
HansNewbie: Noted.

PR #170, opened 2016-06-06
Implement ViewableModel -- Model dual layer #121

### LINKED ISSUE #121: Implement ViewableModel -- Model dual layer for Person objects

YijinL:
Resolves #121 

Should work in place with existing components such that absolutely minimal changes are required on their end.

The EditInProgress etc. refactoring will be done in another PR.

Remaining:

- [x] merge BackingAddressBook with AddressBook
- [x] new subpackages at the domian object level (eg `...model.datatypes.person.*`)
- [x] merge ObservableX with ReadableX as ReadOnlyX
- [x] add ReadOnlyX interfaces to VAB and AB
- [x] write unmodifiable wrapper classes for FX Properties so ReadOnlyX is truly read only


src/main/java/address/model/datatypes/ViewablePerson.java
```diff
@@ -0,0 +1,146 @@
...
+import javafx.collections.ListChangeListener;
+import javafx.collections.ObservableList;
+
+import java.time.LocalDate;
+
+/**
+ * Extends {@link Viewable} for the Person domain object.
+ * @see Viewable
+ */
+public class ViewablePerson extends Viewable<Person> implements ObservableViewablePerson {
```
YijinL: Does not extend `Person` because it should exclude `UniqueData` superclass guarantees.

YijinL:
Completed, requesting review.

src/main/java/address/model/BackingAddressBook.java
```diff
@@ -0,0 +1,74 @@
...
+
+import javafx.collections.FXCollections;
+import javafx.collections.ObservableList;
+
+import java.util.Optional;
+
+/**
+ *
+ */
+class BackingAddressBook implements Model {
```
damithc: Any suggestions for a better name?

src/main/java/address/model/BackingAddressBook.java
```diff
@@ -0,0 +1,74 @@
...
+
+    final ObservableList<Person> allPersons;
+    final ObservableList<Tag> allTags;
+
+    {
```
damithc: add the keyword `static` so that it doesn't look like a violation of our brace placement convention?

YijinL: This is the instance initializer block that runs right before constructor calls; there cant be anything before the leading brace.

src/main/java/address/model/BackingAddressBook.java
YijinL: The `Model` and `VisibleModel` interfaces actually only involve retrieval of the internal observablelists. 

I could use those names for the implementing classes instead and rename the interfaces to `ObservableModel` and `ObservableVisibleModel` respectively

YijinL:
Should I write the unit tests for model before or after merging?
If after, should I write it before or after implementing the `EditInProgress`-type command objects?

src/main/java/address/model/BackingAddressBook.java
damithc: I meant a better term for `Backing`

damithc:
I think unit tests can wait until we have a fully working version of the model. In the meantime you can have some high-level tests to check the core functionality of component as a whole, which can serve as a smoke test.

src/main/java/address/browser/BrowserManager.java
```diff
@@ -2,8 +2,11 @@
 
 import address.events.EventManager;
 import address.events.LocalModelChangedEvent;
+import address.model.datatypes.ReadableViewablePerson;
 import address.model.datatypes.Person;
 
+import address.model.datatypes.ReadablePerson;
+import address.model.datatypes.ReadableViewablePerson;
```
HansNewbie: Unnecessary imports.

src/main/java/address/controller/PersonCardController.java
```diff
@@ -126,17 +127,14 @@ protected String computeValue() {
      * Involves making an internet connection with the image hosting server.
      */
     private void setProfileImage() {
-        Optional<String> profileImageUrl = person.getGithubProfilePicUrl();
-        if (profileImageUrl.isPresent()){
-            new Thread(() -> {
-                Image image = new Image(profileImageUrl.get());
+        new Thread(() -> {
+                Image image = new Image(person.githubProfilePicUrl());
```
HansNewbie: What happens if the person has no Github URL? Won't this fail, as opposed to the original implementation? It could be a malformed URL, e.g. `github.com/<empty>.png`.

src/main/java/address/controller/PersonOverviewController.java
```diff
@@ -72,14 +74,14 @@ public void setConnections(MainController mainController, ModelManager modelMana
...
     @FXML
     private void handleDeletePerson() {
-        int selectedIndex = personList.getSelectionModel().getSelectedIndex();
+        int selectedIndex = personListView.getSelectionModel().getSelectedIndex();
         if (selectedIndex >= 0) {
-            Person tmpPerson = personList.getItems().get(selectedIndex);
-            mainController.getStatusBarHeaderController().postStatus(
-                                                    new PersonDeletedStatus(tmpPerson));
-            personList.getItems().get(selectedIndex).setIsDeleted(true);
+            ObservableViewablePerson target = personListView.getItems().get(selectedIndex);
```
HansNewbie: Variable name `target` is confusing. Target of what; deletion? Maybe `personToBeDeleted` or something more descriptive.

src/main/java/address/controller/PersonOverviewController.java
```diff
@@ -93,14 +95,14 @@ private void handleDeletePerson() {
      */
     @FXML
     private void handleNewPerson() {
-        Optional<Person> newPerson = Optional.of(new Person());
+        Optional<ReadablePerson> prevInput = Optional.of(new Person());
```
HansNewbie: Variable name `prevInput` seems misleading. Maybe `userCreatedPerson` or something more descriptive as well.

src/main/java/address/model/datatypes/ViewablePerson.java
```diff
@@ -0,0 +1,214 @@
...
+    @Override
+    public BooleanProperty isEditedProperty() {
+        return isEdited;
+    }
+
+    public void setIsEdited(boolean isEdited) {
+        this.isEdited.set(isEdited);
+    }
+
+
```
YijinL: Will add comments explaining the state properties after I finalise the "visible state" implementation in the `EditInProgress` PR

src/main/java/address/controller/PersonOverviewController.java
YijinL: I'm also using this variable to store the previous input data to repopulate the dialog if the dialog previously closed with a warning. How does `latestPersonData` sound?

src/main/java/address/model/datatypes/ReadableViewablePerson.java
```diff
@@ -0,0 +1,18 @@
...
+import address.model.ModelManager;
+import javafx.collections.ObservableList;
+
+import java.util.Collection;
+import java.util.function.Supplier;
+
+/**
+ * Same purpose as {@link ReadablePerson}, extended with additional status data viewable by user.
+ */
+public interface ReadableViewablePerson extends ReadablePerson {
```
HansNewbie: I think `ViewableReadablePerson` is a better name. `ReadableViewablePerson` and `ReadablePerson` suggest that we have both `ViewablePerson` and `Person` that will be processed separately. While in this case, we have `ReadablePerson` with `Viewable` property, so `ViewableReadablePerson` suggests that the base object is `ReadablePerson`.

src/main/java/address/controller/PersonOverviewController.java
YijinL: Changing to `deleteTarget`, so if we ever rename Person we dont have to change the name again.

src/main/java/address/controller/PersonOverviewController.java
HansNewbie: Sure.

HansNewbie: Maybe `lastPersonData` is better. `latest` suggests there is an update or something.

src/main/java/address/controller/PersonOverviewController.java
YijinL: deleteTarget?

src/main/java/address/controller/PersonOverviewController.java
YijinL: latestInputPerson? I'm also using it to store the previous input so if the eg. the user tries to add duplicate person we can warn them, then **return them to the dialog and restore their last input**

src/main/java/address/controller/PersonCardController.java
YijinL: reverted

Issue #175, opened 2016-06-06
Clean up storage manager


damithc:
as per title

HansNewbie:
Seems that the comment in the PR was not working. This issue was resolved in #176.

PR #176, opened 2016-06-06
Clean up storage manager #175

damithc:
* Made dependencies injectable
* Separated 'Manager' tasks from other tasks by adding a `XmlFileStorage` class
* Trimmed `XmlFileHelper` class
* Stopped StorageManager from raising `SaveRequestEvent` (this class should respond to the event rather than raise it). Likely to break something somewhere, but hopefully it will get sorted out later.

### DIFF
--- a/src/main/java/address/model/AddressBook.java
+++ b/src/main/java/address/model/AddressBook.java
@@ -22,6 +22,15 @@ public class AddressBook {
     private List<Person> persons = new ArrayList<>();
     private List<Tag> tags = new ArrayList<>();
 
+    //TODO: remove this empty constructor
+    public AddressBook (){
+    }
+
+    public AddressBook (List<Person> persons, List<Tag> tags){
+        this.persons = persons;
+        this.tags = tags;
+    }
+
     @XmlElement(name = "persons")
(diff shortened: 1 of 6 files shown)


src/main/java/address/model/AddressBook.java
```diff
@@ -22,6 +22,15 @@
     private List<Person> persons = new ArrayList<>();
     private List<Tag> tags = new ArrayList<>();
 
+    //TODO: remove this empty constructor
+    public AddressBook (){
```
HansNewbie: I don't think it's possible to remove empty constructor for any classes to be serialized. All serializers (JAXB, GSON and Jackson) need an empty constructor to be present for deserialization.

Jackson allows another way to bypass this, but many annotations must be added to the constructor with parameters which will clutter the constructor signature. [There is a way to bypass this as well but it requires changing the compiler option](https://manosnikolaidis.wordpress.com/2015/08/25/jackson-without-annotations/).

damithc:
@HansNewbie shall I merge this one?

HansNewbie:
Good to merge.

HansNewbie:
This PR will fix #175

PR #182, opened 2016-06-06
Added a base class for Manager classes

damithc:
Added a base class for `StorageManger`. 
Can be used by other Manger classes later.
I'm creating PR without an issue to reduce overhead.


damithc:
I wasn't sure about that myself, but seems to work. Yes, need to keep a lookout in case this makes the events to go to the wrong place.

PR #193, opened 2016-06-08
Add support for global hotkeys 

damithc:
Added two global hotkeys using jkeymaster
`Ctrl+Shift+X` for maximizing window 
`Ctrl+Alt+X` for minimizing window

### DIFF
--- a/src/main/java/address/events/MinimizeAppRequestEvent.java
+++ b/src/main/java/address/events/MinimizeAppRequestEvent.java
@@ -1,0 +1,11 @@
+package address.events;
+
+/**
+ * Represents a user request to minimize the app window
+ */
+public class MinimizeAppRequestEvent extends BaseEvent {
+    @Override
+    public String toString() {
+        return "Request to maximize app window";
+    }
+}
--- a/src/main/java/address/events/MinimizeAppRequestEvent.java
+++ b/src/main/java/address/events/MinimizeAppRequestEvent.java
@@ -6,6 +6,6 @@ package address.events;
 public class MinimizeAppRequestEvent extends BaseEvent {
     @Override
     public String toString() {
-        return "Request to maximize app window";
+        return "Request to minimize app window";
     }
 }
--- a/src/main/java/address/shortcuts/Bindings.java
+++ b/src/main/java/address/shortcuts/Bindings.java
@@ -14,21 +14,25 @@ import java.util.Optional;
  * Contains the mapping of shortcuts and the corresponding event to raise
  */
 public class Bindings {
-    private static List<Shortcut> shortcuts = new ArrayList<>();
-    private static List<GlobalHotkey> hotkeys = new ArrayList<>();
+    private List<Shortcut> shortcuts = new ArrayList<>();
+    private List<GlobalHotkey> hotkeys = new ArrayList<>();
 
     /* shortcuts in alphabetical order of names */
-    public static final List<GlobalHotkey> HOTKEY_APP_MINIMIZE = new ArrayList<>();
-    public static final List<GlobalHotkey> HOTKEY_APP_MAXIMIZE = new ArrayList<>();
-    public static final Shortcut SHORTCUT_FILE_NEW;
-    public static final Shortcut SHORTCUT_FILE_OPEN;
-    public static final Shortcut SHORTCUT_FILE_SAVE;
-    public static final Shortcut SHORTCUT_FILE_SAVE_AS;
-    public static final Shortcut SHORTCUT_LIST_ENTER;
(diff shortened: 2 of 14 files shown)


damithc:
@HansNewbie can test on Mac?

HansNewbie:
Alright. Trying now. There is a problem with the gradle not able to resolve the jkeymaster dependency.

src/main/java/address/shortcuts/ShortcutsManager.java
```diff
@@ -41,6 +49,9 @@
 
         SHORTCUT_FILE_SAVE_AS = setShortcut(KeyCode.S, KeyCombination.CONTROL_DOWN, KeyCombination.ALT_DOWN);
 
+        HOTKEY_APP_MINIMIZE = setHotkey("control alt X", new MinimizeAppRequestEvent());
+        HOTKEY_APP_MAXIMIZE = setHotkey("control shift X", new MaximizeAppRequestEvent());
```
HansNewbie: For Mac, `command` button is preferred as shortcut button instead of `control`. That can be achieved by using `"meta alt X"` for minimize case (tested).

damithc:
Restructured the component further

src/main/java/address/events/MinimizeAppRequestEvent.java
```diff
@@ -0,0 +1,11 @@
+package address.events;
+
+/**
+ * Represents a user request to minimize the app window
+ */
+public class MinimizeAppRequestEvent extends BaseEvent {
+    @Override
+    public String toString() {
+        return "Request to maximize app window";
```
HansNewbie: Message should be `minimize`.

src/main/java/address/shortcuts/Bindings.java
```diff
@@ -0,0 +1,161 @@
...
+
+    /**
+     * List of keyboard shortcuts.
+     */
+    private List<Shortcut> shortcuts = new ArrayList<>();
+
+
+    /* Global hotkeys in alphabetical order of name */
+    public List<GlobalHotkey> HOTKEY_APP_MINIMIZE = new ArrayList<>();
+    public List<GlobalHotkey> HOTKEY_APP_MAXIMIZE = new ArrayList<>();
```
HansNewbie: *Minor -* the comment mentions alphabetical order, but that means `MAXIMIZE` should come before `MINIMIZE`

src/main/java/address/shortcuts/Bindings.java
```diff
@@ -0,0 +1,161 @@
...
+
+        ACCELERATOR_PERSON_EDIT = setAccelerator(KeyCode.E);
+
+        ACCELERATOR_FILE_NEW = setAccelerator(KeyCode.N, KeyCombination.CONTROL_DOWN);
+
+        ACCELERATOR_FILE_OPEN = setAccelerator(KeyCode.O, KeyCombination.CONTROL_DOWN);
+
+        ACCELERATOR_FILE_SAVE = setAccelerator(KeyCode.S, KeyCombination.CONTROL_DOWN);
+
+        ACCELERATOR_FILE_SAVE_AS = setAccelerator(KeyCode.S, KeyCombination.CONTROL_DOWN, KeyCombination.ALT_DOWN);
```
HansNewbie: For all `CONTROL_DOWN`, we can try `SHORTCUT_DOWN` which *might* be correctly mapped to `Cmd` on Mac OS X and `Ctrl` on Windows (even if run on Macbook), as compared to `META_DOWN`. Otherwise, we can keep it to `CONTROL_DOWN`.

src/main/java/address/shortcuts/ShortcutsManager.java
```diff
@@ -1,138 +1,56 @@
...
 
-    /* shortcuts in alphabetical order of names */
-    public static final KeyCombination SHORTCUT_FILE_NEW;
-    public static final KeyCombination SHORTCUT_FILE_OPEN;
-    public static final KeyCombination SHORTCUT_FILE_SAVE;
-    public static final KeyCombination SHORTCUT_FILE_SAVE_AS;
-    public static final KeyCombination SHORTCUT_LIST_ENTER;
-    public static final KeyCombination SHORTCUT_PERSON_DELETE;
-    public static final KeyCombination SHORTCUT_PERSON_EDIT;
+    public static Bindings BINDINGS;
```
HansNewbie: Should this be `final`?

damithc: I omitted `final` because it is not initialized at that point.

HansNewbie: Right. I didn't realize it is `static`.

PR #195, opened 2016-06-09
Remove EventManager dependency from sub classes of ComponentManager

damithc:
Child classes don't have to refer to `EventManger` now. The dependency is managed my the parent class `ComponentManager`


HansNewbie:
Looks good.

PR #209, opened 2016-06-11
Unified different types of key bindings

damithc:
Unified the 4 types of key bindings under a parent class `KeyBinding`
Renamed component to `keybindings`


damithc:
Going to merge this. you can still review/comment on the code later and I'll revise accordingly in future PRs.

PR #213, opened 2016-06-13
Wrap logger so that we can have customised message formats #211

### LINKED ISSUE #211: Wrap logger so that we can have customised message formats

m133225:
Fixes #211.


m133225:
Will also be fixing logging level not working correctly after reading from config.

PR #230, opened 2016-06-16
Increase usage of logger #227

### LINKED ISSUE #227: Increase usage of logger
to allow us to debug solely using log messages.

m133225:
Fixes #227.

- [x] Add log statements (just to log more extensively, not planning to cover everything)
- [x] Use JSON logger config instead


damithc:
Was xml config files (as opposed to json, properties, etc.) a conscious choice?

m133225:
For the log message formatting (in log4j2.xml), no. Used XML because it was easier for me to read, and it was the commonly used format in the guide. Should we use JSON instead?

On a side note, `ini` was the conscious choice for `config.ini` instead of JSON simply due to readability/understandability (by a layman).

PR #243, opened 2016-06-17
Improve drag and drop #240

### LINKED ISSUE #240: Improve drag and drop
Improve drag and drop experience by showing indicator on where will the cell be dropped. Allow dropping either on top or below the selected cell.

#228

yl-coder:
Fixes #240 

Summary on what happen in this PR, in case we need to refer back in future.

Fix three major bug.
1) JxBrowser loadUrl: this method is called asynchronously. So the url you passed in may not be reflected on the getUrl() method. If loadUrl is called consecutively. For e.g. thrice, calling getURL may result getting any of the 3 method calls.
Solution: created a custom class JxBrowser. Added an additional field call originUrl. It is the url that the browser is last tasked to load.

2) Multiple instance of jxBrowser browser. Multiple browser instance will share the same common resources, so if no configuration is set. Deadlocks may occurred.
Solution: configure different browser instance to access its individual resources.

3) Calling jxBrowser API in non-UI thread will cause deadlock
Solution: use Platform.runLater if need to access jxBrowser API at non-UI thread.

Special note:
When UI thread is busy(e.g. drag and drop operation), it may increase deadlock chances with jxBrowser threads(Maybe, just assumption for now).


src/main/java/address/util/OrderedList.java
```diff
@@ -0,0 +1,73 @@
...
+    private ObservableList<T> orderedList;
+
+    /**
+     * Creates a new Transformation list wrapped around the source list.
+     *
+     * @param source the wrapped list
+     */
+    public OrderedList(ObservableList<T> source) {
+        super(source);
+        orderedList = FXCollections.observableArrayList();
```
HansNewbie: Won't this create empty list? Then `source` is not used.

src/main/java/address/browser/HyperBrowser.java
```diff
@@ -74,7 +76,8 @@ private void initialiseHyperBrowser(){
         for (int i = 0; i < noOfPages; i++){
             EmbeddedBrowser browser;
             if (browserType == FULL_FEATURE_BROWSER){
-                browser = new JxBrowserAdapter(new Browser());
+                BrowserContext context = new BrowserContext(new BrowserContextParams("tmpTab" + i));
```
HansNewbie: What's the context for? The documentation states for data directory such as cookies. If that's the case, we shouldn't separate them as the same user will be using the different tabs and it's easier if the different tabs use the same cookies.

src/main/java/address/browser/jxbrowser/JxBrowser.java
```diff
@@ -0,0 +1,43 @@
...
+import com.teamdev.jxbrowser.chromium.BrowserContext;
+import com.teamdev.jxbrowser.chromium.BrowserType;
+
+/**
+ * A JxBrowser instance with the ability to know the URL it was tasked to load.
+ */
+public class JxBrowser extends Browser{
+
+
+
```
HansNewbie: unnecessary new lines.

src/main/java/address/util/OrderedList.java
yl-coder: The list will be updated in the sourceChanged method. at first the source is empty also.

src/main/java/address/browser/HyperBrowser.java
yl-coder: I agree, but previously, there is a dead lock on jxBrowser getUrl method when this is not set. After I changed this, the problem seems to be gone.

src/main/java/address/browser/jxbrowser/JxBrowser.java
yl-coder: ok fixed

src/main/java/address/ui/PersonListViewCell.java
```diff
@@ -72,8 +96,78 @@ public PersonListViewCell(ReorderedList sortedList) {
...
+    /**
+     * Scrolls up or down if pointer reaches the edge(top and bottom) of the listview.
+     * @param event
+     */
+    private void scrollIfNecessary(DragEvent event) {
+        double maxY = getListView().localToScene(getListView().getBoundsInLocal()).getMaxY();
+        double minY = getListView().localToScene(getListView().getBoundsInLocal()).getMinY();
+        Optional<VirtualScrollBar> scrollbar = FxViewUtil.getScrollBarFromListView(getListView());
+        if (scrollbar.isPresent()) {
+            int scrollArea = 15;
```
HansNewbie: What's this `15`? If it's a magic number, abstract as constant.

yl-coder: ok noted. fixed

src/main/java/address/util/OrderedList.java
```diff
@@ -0,0 +1,73 @@
...
+    public void moveElement(int from, int to) {
+        if (from < to) {
+            T tmpPerson = orderedList.get(from);
+            orderedList.add(to, tmpPerson);
+            orderedList.remove(from);
+        } else if (from > to) {
+            T tmpPerson = orderedList.remove(from);
+            orderedList.add(to, tmpPerson);
+        }
+    }
```
HansNewbie: Is the different treatment necessary? i.e., can't both use the same call? If cannot, put some comments.

src/main/java/address/util/OrderedList.java
HansNewbie: Maybe put comment that source is empty originally. If that's the case, is the super call necessary?

HansNewbie:
No more comments.

src/main/java/address/util/OrderedList.java
yl-coder: removing the element below the index where it needs to move to will decrease the index of "to", but if the element is above the "to" index. Then removing doesn't have any effect on the index.

src/main/java/address/util/OrderedList.java
yl-coder: It is needed by the class.

src/main/java/address/util/OrderedList.java
HansNewbie: Noted. Comments would be helpful.

src/main/java/address/util/OrderedList.java
HansNewbie: Okay.

yl-coder: copy the empty list to be safe.
orderedList = FXCollections.observableArrayList(source);
Since it may not be empty in other scenario if we happen to use this OrderedList on other application

PR #249, opened 2016-06-19
Reduced exposure of the Bindings class

damithc:
Made `Bindings` class private
Moved the test class to the same package as SUT (to take advantage of `package` access)

### DIFF
--- a/src/main/java/address/controller/PersonOverviewController.java
+++ b/src/main/java/address/controller/PersonOverviewController.java
@@ -167,10 +167,10 @@ public class PersonOverviewController {
         final ContextMenu contextMenu = new ContextMenu();
 
         MenuItem editMenuItem = new MenuItem("Edit");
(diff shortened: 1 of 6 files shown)


damithc:
@m133225 it looks like it is better to keep the test class as same as the SUT. That way, tests can access SUT members that are of `package` access. See if there is any downside to it.

m133225:
I don't think they are many downsides to it.

We could put the unit tests and the gui unit tests in same packages as production code; the disadvantage is that we cannot easily differentiate gui and non-gui unit tests, but that shouldn't affect us too much, since there will not be many gui unit tests in the first place.

For the integration gui tests, we can always put them in a new **guitests** package which can be placed in the `test/java` folder.

damithc:
I moved the test classes back to the original location as I don't need package access just yet. We can do a mass move later.

PR #258, opened 2016-06-20
Extract searching of tags into a separate dialog #256

### LINKED ISSUE #256: Extract searching of tags into a separate dialog
From person edit view.

m133225:
Fixes #256


m133225:
Updated dialog(s):

PR #262, opened 2016-06-21
Tip of the day #260

### LINKED ISSUE #260: Tip of the day
Tip of the day should not block the focus of the application so user can try out the tip (say keyboard shortcut) without needing to click the main application.

HansNewbie:
Fixes #260

### DIFF
--- a/src/main/java/address/controller/MainController.java
+++ b/src/main/java/address/controller/MainController.java
@@ -10,11 +10,7 @@ import address.model.datatypes.person.ReadOnlyPerson;
 import address.model.datatypes.tag.Tag;
 import address.util.*;
 import com.google.common.eventbus.Subscribe;
-import javafx.animation.KeyFrame;
-import javafx.animation.KeyValue;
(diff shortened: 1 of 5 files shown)


HansNewbie:
@damithc any comments on this?

What has been done:
- Tip of the day (TOTD) will slide in from top right at startup.
- TOTD won't block focus and won't close if user take action. Demo is a bit bad since I was trying out global hotkey, but with tip shown, I can still change filter and use keyboard shortcut on person cards.
- TOTD is closed by clicking on it.
- Tips will be stored as JSON in `resources/tipsOfTheDay.json`.

Demo:
tip of the day

Yet to be done (possibly on separate PRs):
- When to show tip. We can implement timer or delay on when to show TOTD in `TipOfTheDayController`.
- Styling of tips.
- Closing tip when user take action (I am not sure how we would like to do this).

damithc:
Good enough for a first version. We can refine later.

PR #265, opened 2016-06-21
Made gui tests more configurable

damithc:
This PR pushes toward making tests (gui tests in particular) more configurable and less coupled from production code/data. e.g. running tests should not mess up the data of the app.
Some things done:
* Cleaned up MainApp code a bit
* Extracted a UserPrefs object
* Removed loading sample data (using menu) from the beginning of GUI tests

Some more work to do though.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ lib/*
 /*.xml
 update/
 src/main/resources/updater/jarUpdater.jar
+src/test/data/sandbox/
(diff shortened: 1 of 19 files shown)


damithc:
@m133225 see if the code is ok.

m133225:
Looks ok.

Are we planning to have multiple users? If so, we should consider implementing user-bound prefs in the near future.

damithc:
User-bound prefs is nice to have, but probably we can do without in addressbook and also in HT.

PR #271, opened 2016-06-22
Move file location to footer #233

### LINKED ISSUE #233: Move file location to footer
This can be moved to the footer (periodic status bar). This space is for the action status bar.
image

yl-coder:
Fixes #233 

image

### DIFF
--- a/src/main/java/address/controller/MainController.java
+++ b/src/main/java/address/controller/MainController.java
@@ -146,13 +146,12 @@ public class MainController {
 
     private void showHeaderStatusBar() {
         statusBarHeaderController = new StatusBarHeaderController();
-        AnchorPane sbPlaceHolder = (AnchorPane) rootLayout.lookup("#footerStatusbarPlaceHolder");
-        if (sbPlaceHolder == null) {
-            logger.warn("Error finding layout node");
-            showAlertDialogAndWait(AlertType.ERROR, "Error loading layout resources", "", "");
-        }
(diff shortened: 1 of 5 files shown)


src/main/java/address/controller/MainController.java
```diff
@@ -146,7 +146,13 @@ public StatusBarHeaderController getStatusBarHeaderController() {
 
     private void showHeaderStatusBar() {
         statusBarHeaderController = new StatusBarHeaderController();
-        rootLayout.getChildren().add(2, statusBarHeaderController.getFooterStatusBarView());
+        AnchorPane sbPlaceHolder = (AnchorPane) rootLayout.lookup("#footerStatusbarPlaceHolder");
```
YijinL: Maybe declare as object field with `@FXML` so the loader auto lookups for us?

src/main/java/address/controller/MainController.java
```diff
@@ -146,7 +146,13 @@ public StatusBarHeaderController getStatusBarHeaderController() {
...
+        if (sbPlaceHolder == null) {
+            logger.warn("Error finding layout node");
+            showAlertDialogAndWait(AlertType.ERROR, "Error loading layout resources", "", "");
+        }
```
damithc: Is this the correct error handling. Looks like this should be an assertion failure that should abort the app launch.

yl-coder: Yup right, changed it. 

I guess we need to change it to assertion for error in loading FXML also.

Will create another PR for changing the FXML one.

PR #272, opened 2016-06-22
Convert preference storage to json format #266

### LINKED ISSUE #266: Convert preference storage to json format
This way, it will be more visible and easier to reduce configuration leaks between tests.
Although java has a Preferences API, it doesn't seem to be well developed beyond simply key/value pairs.

HansNewbie:
Fixes #266


damithc:
Error handling is half baked, but may be ok because we are going to migrate this later.

PR #275, opened 2016-06-22
Treat UserPrefs as Model

HansNewbie:


HansNewbie:
@damithc is this a good implementation of what we discussed?

I will move the writing of preferences file to `StorageManager` if this is accepted.

src/main/java/address/controller/RootLayoutController.java
```diff
@@ -54,15 +51,14 @@ public RootLayoutController() {
     }
 
     @FXML
-    private void initialize() {
-        updateSaveLocationDisplay();
-    }
+    private void initialize() {}
```
HansNewbie: Can't call `updateSaveLocationDisplay` here because the `modelManager` is not passed in yet in `setConnections`. Moved it to `setConnections` instead. Might be better to abstract it out into some sort of initialize method that is outside of FXML initialize.

src/main/java/address/prefs/UserPrefs.java
```diff
@@ -29,6 +31,12 @@ public void setSaveLocation(String saveLocation) {
         }
 
         writePrefToFile();
+
+        EventManager.getInstance().post(new SaveLocationChangedEvent(new File(saveLocation)));
```
HansNewbie: If we are to move this to a manager, we need to wrap `UserPrefs` methods into `ModelManager` methods and post the event there.

src/main/java/address/controller/RootLayoutController.java
HansNewbie: No longer relevant due to #271

HansNewbie:
Yet to move the `UserPrefs` to `model` package.

src/main/java/address/controller/RootLayoutController.java
```diff
@@ -82,7 +75,7 @@ private FileChooser getXmlFileChooser() {
     @FXML
     private void handleNew() {
         logger.debug("Wiping current model data.");
-        PrefsManager.getInstance().clearSaveLocation();
+        modelManager.getPrefs().clearSaveLocation();
```
damithc: Edits to prefs should go through a method in ModelManger, not through the prefs object (which should be read only)

HansNewbie: Will do. How about for components that takes in `UserPrefs` in constructor? Should I make an interface that allows them to read only?

HansNewbie:
@damithc to create that read-only access to `UserPrefs` using packaging, I need to put `UserPrefs` in the same package as `ModelManager` (that is, `model`) and not in `model.datatypes` package. Is that alright?

damithc:
Either that or if that makes things messy you go for the 'ReadOnly interface' option.

HansNewbie:
I will leave it at that for now as it does not seem messy.

HansNewbie:
@damithc I will merge if no more comments

src/main/java/address/events/LoadPrefsRequestEvent.java
```diff
@@ -0,0 +1,20 @@
...
+
+    public LoadPrefsRequestEvent(File file) {
+        this.file = file;
+    }
+
+    @Override
+    public String toString(){
+        return "from " + file.getAbsolutePath();
+    }
+}
```
damithc: If loading preferences from the file happens only at the startup, we don't need an event for it. The MainApp can load the UserPrefs object using StorageManager (a static method?) and pass it to the constructor of ModelManager.

src/main/java/address/controller/RootLayoutController.java
```diff
@@ -82,7 +75,7 @@ private FileChooser getXmlFileChooser() {
     @FXML
     private void handleNew() {
         logger.debug("Wiping current model data.");
-        PrefsManager.getInstance().clearSaveLocation();
+        modelManager.clearPrefsSaveLocation();
```
damithc: Not related to this PR, but why the save location is resest here? The purpose of this menu option is to wipeout the data, not the location where it is saved, right?

src/main/java/address/storage/StorageManager.java
```diff
@@ -22,15 +23,13 @@
 public class StorageManager extends ComponentManager {
     private static final AppLogger logger = LoggerManager.getLogger(StorageManager.class);
 
-    private ModelManager modelManager;
-    private UserPrefs userPrefs;
+    public static final File DEFAULT_USER_PREF_FILE = new File("pref");
```
damithc: Should be `preferences.json`?

damithc:
No more comments from me for this version.

src/main/java/address/storage/StorageManager.java
HansNewbie: Noted.

src/main/java/address/controller/RootLayoutController.java
HansNewbie: I am not sure. Introduced in #91. I am guessing when you create a new addressbook you want to clear the save file so it won't erase the current addressbook file.

Should I remove it?

src/main/java/address/events/LoadPrefsRequestEvent.java
HansNewbie: Noted.

HansNewbie: Seems that's impossible, though. Because `StorageManager` needs `ModelManager` to be passed to its constructor. I can still load the prefs from `StorageManager` by calling `ModelManager.updatePrefsFromExternalData`, if that's fine with you.

Otherwise, `ModelManager` will need to update addressbook using event. (currently `StorageManager` will call `ModelManager.updateUsingExternalData` upon handling load event).

damithc: Use a static method?

HansNewbie: Oh right. Missed that. Will do.

pref
```diff
@@ -0,0 +1,3 @@
+{
+  "saveLocation" : "/Users/hans/Documents/Workspace/addressbook/src/test/data/sandbox/Untitled.xml"
+}
```
damithc: Should this file be here?

HansNewbie: My mistake. Put in while rebasing. Will remove.

src/main/java/address/controller/RootLayoutController.java
damithc: I see. OK, keep it.

damithc: May be add a comment?

HansNewbie: Sure.

src/main/java/address/storage/StorageManager.java
```diff
@@ -136,11 +136,49 @@ public void saveDataToFile(File file, ReadOnlyAddressBook data){
...
+        return prefs;
+    }
+
+    /**
      * Loads the data from the local data file (based on user preferences).
      */
     public void start() {
         logger.info("Starting storage manager.");
-        loadDataFromFile(userPrefs.getSaveLocation());
+        loadDataFromFile(modelManager.getPrefs().getSaveLocation());
```
damithc: It's better for StorageManger to keep a copy of preferences rather than get it from model manager. Storage location is StorageManager's business, not ModelManager's.

damithc:
Added some comments

src/main/java/address/storage/StorageManager.java
HansNewbie: Sure. Should I:
- pass the `UserPrefs` in the constructor parameter, i.e. `StorageManager(ModelManager modelManager, UserPrefs prefs)`, or
- just load it inside constructor since we can get it from `ModelManager`, i.e.
```java
public StorageManager(ModelManager modelManager) {
  this.modelManager = modelManager;
  this.prefs = modelManager.getPrefs();
  .
  .
  .
}
```

src/main/java/address/storage/StorageManager.java
```diff
@@ -21,13 +21,15 @@
...
     private ModelManager modelManager;
-    private UserPrefs userPrefs;
+    private UserPrefs prefs;
 
-    public StorageManager(ModelManager modelManager, UserPrefs userPrefs) {
+    public StorageManager(ModelManager modelManager) {
         super();
         this.modelManager = modelManager;
-        this.userPrefs = userPrefs;
+        this.prefs = modelManager.getPrefs();
```
HansNewbie: Is this fine? Or should I pass `prefs` in constructor?

src/main/java/address/storage/StorageManager.java
damithc: I prefer to pass it as parameter rather than get it from model manager. It makes the dependency more obvious.

HansNewbie: Sure.

HansNewbie:
Ready for review.

damithc:
No more comments from me

PR #277, opened 2016-06-22
Config should not be a singleton #273

### LINKED ISSUE #273: Config should not be a singleton

m133225:
Fixes #273


src/main/java/address/MainApp.java
```diff
@@ -41,34 +41,38 @@
...
     protected void initPrefs() {
         // For sub classes to override
     }
 
     protected void initComponents() {
-        modelManager = new ModelManager();
         storageManager = new StorageManager(modelManager, PrefsManager.getInstance().getPrefs());
-        mainController = new MainController(this, modelManager);
-        syncManager = new SyncManager();
+        initConfig();
```
damithc: called twice

src/main/java/address/MainApp.java
```diff
@@ -41,34 +41,38 @@
...
     public void init() throws Exception {
         logger.info("Initializing app ...");
         super.init();
+        initComponents();
         initConfig();
-        Config.setConfig(Config.getConfig());
         initPrefs();
         BrowserManager.initializeBrowser();
         //TODO: should this be here? looks out of place
-        initComponents();
```
damithc: I was expecting to see this order:
```
initConfig();
initPrefs();
initComponents();
```

damithc:
Added two comments. Probably you are not yet done with the code, so ignore my comments if not applicable.

src/main/java/address/MainApp.java
m133225: There's a problem with this order.
- Initializing the components also means initializing the storage manager, which is responsible for reading from the config file. i.e. the config will get overwritten by components initialization in TestApp.

Either way, there's some weird Exception being thrown in the start method at the moment. Will need to figure that out first.

m133225: Actually I guess we could check if it is instantiated in the first place before overriding with the read configs, but it just seems weird if the initConfig method is empty in the MainApp.

m133225:
Will move Config out of StorageManager. Don't think Config should be dependent on StorageManager since all other components (including StorageManager) may depend on values in Config.

PR #279, opened 2016-06-22
Made SyncManager extend ComponentManager

damithc:


damithc:
@m133225 for your review and merging

src/main/java/address/sync/SyncManager.java
```diff
@@ -28,18 +29,16 @@
     private RemoteManager remoteManager;
 
     public SyncManager() {
-        activeAddressBook = Optional.empty();
-        scheduler = Executors.newScheduledThreadPool(1);
-        requestExecutor = Executors.newCachedThreadPool();
-        EventManager.getInstance().registerHandler(this);
```
m133225: I think this shouldn't be removed? Needed for `SyncManager` to handle certain events.

HansNewbie: It's in `super()` call of `ComponentManager`.

damithc:
@m133225 go ahead and merge if ok.

PR #282, opened 2016-06-23
Tooltip footer, relocation of browsermanager init, remove version from footer status bar

### LINKED ISSUE #278: Relocation of initComponents()
The creation of browserManager instance need to be after the stage is created. As JavaFX webview needs it. 

Currently it is created before the stage is created.

Any suggestions guys, on where to move to?

```

    @Override
    public void init() throws Exception {
        logger.info("Initializing app ...");
        super.init();
        initConfig();
        Config.setConfig(Config.getConfig());
        initPrefs();
        BrowserManager.initializeBrowser();
        //TODO: should this be here? looks out of place
        initComponents();
    }
```

yl-coder:
Fixes #278

### DIFF
--- a/src/main/java/address/controller/StatusBarFooterController.java
+++ b/src/main/java/address/controller/StatusBarFooterController.java
@@ -32,11 +32,11 @@ public class StatusBarFooterController {
 
     private final long updateIntervalInSecs;
 
-    private final Label saveLocText;
+    private final Label saveLocationLabel;
 
     public StatusBarFooterController() {
         EventManager.getInstance().registerHandler(this);
-        this.saveLocText = new Label();
+        this.saveLocationLabel = new Label();
         Config config = Config.getConfig();
         updateIntervalInSecs = (int) DateTimeUtil.millisecsToSecs(config.updateInterval);
         timer = new TickingTimer("Sync timer", (int) updateIntervalInSecs,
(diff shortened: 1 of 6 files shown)


m133225:
Can't it be called in `BrowserManager`'s constructor? Didn't realise you were also changing this, will remove in my PR.

yl-coder:
@m133225 The initialise method of browser in mainApp must be called in a non ui. So maybe thats the reason why it is placed there instead at the mainApp.

src/main/java/address/controller/StatusBarFooterController.java
```diff
@@ -55,8 +56,12 @@ private void syncSchedulerOnTimeOut() {
...
         this.syncStatusBar = new StatusBar();
         this.updaterStatusBar = new StatusBar();
-        this.updaterStatusBar.getLeftItems().add(saveLocText);
-
+        saveLocText.setTextAlignment(TextAlignment.LEFT);
+        Tooltip tp = new Tooltip();
+        tp.textProperty().bind(saveLocText.textProperty());
+        saveLocText.setTooltip(tp);
+        this.updaterStatusBar.getRightItems().add(saveLocText);
+        saveLocText.setVisible(false);
```
HansNewbie: Abstract this to a method? Setting up of tooltip here breaks the abstraction level.

HansNewbie:
No more comments.

damithc:
Why did we remove version from status bar?

yl-coder:
Since we have the save location detail in the footer, i guess we can move the version to the about section? Since the user does not need to know about the version all the time.

damithc:
I see. Yes, that's the better place for version. However, we should add some more stuff to the footer to simulate a legit footer showing many different data items.

yl-coder:
Right. Will try to introduce more stuff into the footer.

PR #283, opened 2016-06-23
Autoscroll to bottom broken #276

### LINKED ISSUE #276: Autoscroll to bottom broken
Doesn't seem to work now. Probably something regressed during the update to the model manager.

yl-coder:
Fixes #276

### DIFF
--- a/src/main/java/address/browser/BrowserManager.java
+++ b/src/main/java/address/browser/BrowserManager.java
@@ -2,7 +2,6 @@ package address.browser;
 
 import address.browser.page.GithubProfilePage;
 import hubturbo.embeddedbrowser.BrowserType;
-import hubturbo.embeddedbrowser.EmbeddedBrowserFactory;
 import hubturbo.embeddedbrowser.HyperBrowser;
 import hubturbo.embeddedbrowser.page.Page;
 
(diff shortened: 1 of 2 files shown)


src/main/java/address/browser/BrowserManager.java
```diff
@@ -101,7 +100,8 @@ public synchronized void loadProfilePage(ReadOnlyViewablePerson person) {
         try {
             Page page = hyperBrowser.get().loadUrls(person.profilePageUrl(), listOfFutureUrl);
             GithubProfilePage gPage = new GithubProfilePage(page);
-            gPage.setPageLoadFinishListener(b -> Platform.runLater(() -> gPage.activateAutomateClickingAndScrolling()));
+            gPage.executePageLoadedTasks();
```
HansNewbie: We need this to fix the bug?

yl-coder: Yes, if the page is already loaded, then execute the tasks.

HansNewbie: If the call below (`setPageLoadFinishListener`) is not enough, then sure. No more comments.

PR #290, opened 2016-06-23
Add more qualifiers for better filter functionalities #255

### LINKED ISSUE #255: Add more qualifiers for better filter functionalities

m133225:
Fixes #255 

Will also be fixing filters as a whole since it stopped working at some point.

### DIFF
--- a/pref
+++ b/pref
@@ -1,0 +1,3 @@
+{
+  "saveLocation" : null
+}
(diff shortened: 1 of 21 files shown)


damithc:
Don't add too many qualifiers though, as we don't gain any new knowledge by doing so.

m133225:
How should the interaction be in the following scenario:
1. List is filtered by X:
```
Person 1
Person 2 (hidden by filter)
Person 3
```
2. User reorders the person cards (Drag `Person 3` to above `Person 1`)
```
Person 3
Person 1
Person 2 (hidden by filter)
```
3. 
i. User re-filters using X again
- Should the order be reset to the original order?
- Or should the new order be retained?

ii. User removes X
- Should the order be 1,2,3 or 3,1,2 (Person 2 which matches the filter be added to the end)?

damithc:
Do the easier thing. The reordering feature was just to try to DnD action. It is not meant to be that useful in the current state of the app.

PR #292, opened 2016-06-24
Made app init consistent and phased

damithc:
Made the init stages more natural, incremental.
Also eliminated the settings leak between manual testing and automated tests.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,9 +1,0 @@
-{
-  "appTitle" : "Address App",
-  "updateInterval" : 10000,
-  "simulateUnreliableNetwork" : true,
-  "currentLogLevel" : "DEBUG",
-  "specialLogLevels" : {
-    "SyncManager" : "TRACE"
-  }
-}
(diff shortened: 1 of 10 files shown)


damithc:
@m133225 @YijinL @HansNewbie have a look (this involves config, prefs, and model manager)

config.json
```diff
@@ -1,9 +0,0 @@
-{
-  "appTitle" : "Address App",
-  "updateInterval" : 10000,
-  "simulateUnreliableNetwork" : true,
-  "currentLogLevel" : "DEBUG",
-  "specialLogLevels" : {
-    "SyncManager" : "TRACE"
-  }
-}
```
damithc: No need for this file to be tracked as the default config file is created automatically. Not tracking it allows devs to have their own config.

HansNewbie:
Looks awesome 👍

src/main/java/address/model/ModelManager.java
```diff
@@ -46,7 +46,8 @@
      * AddressBook and its variables should not be null.
      * @param src
      */
-    public ModelManager(AddressBook src, UserPrefs prefs) {
+
```
YijinL: Should there be a new line between the method declaration and it's header comment?

YijinL:
Looks good

src/main/java/address/model/ModelManager.java
damithc: No. got added during conflict resolution. I'll remove it.

PR #297, opened 2016-06-24
Added a test to detect clashes between key sequences and accelerators

damithc:
A safety mechanism to detect clashing accelerators and key sequences


HansNewbie:
Looks good.

PR #299, opened 2016-06-25
Added a UI component to front all ui-related code

damithc:
The UI class will act as the facade for UI code, including the embedded browser.

### DIFF
--- a/src/main/java/address/ui/Ui.java
+++ b/src/main/java/address/ui/Ui.java
@@ -1,0 +1,34 @@
+package address.ui;
+
+import address.MainApp;
+import address.browser.BrowserManager;
+import address.controller.MainController;
+import address.model.ModelManager;
+import address.util.Config;
+import javafx.scene.control.Alert;
+import javafx.stage.Stage;
+
+/**
+ * The UI of the app.
+ */
+public class Ui {
+    MainController mainController;
+
+    public Ui(MainApp mainApp, ModelManager modelManager, Config config){
+        BrowserManager.initializeJxBrowserEnvironment();
+        //TODO: this should not be here because at this point we are not even sure which browser will be used
+        mainController = new MainController(mainApp, modelManager, config);
+    }
+
+    public void start(Stage primaryStage) {
+        mainController.start(primaryStage);
+    }
+
+    public void showAlertDialogAndWait(Alert.AlertType alertType, String alertTitle, String headerText, String contentText) {
+        mainController.showAlertDialogAndWait(alertType, alertTitle, headerText, contentText);
+    }
+
+    public void stop() {
+        mainController.stop();
+    }
+}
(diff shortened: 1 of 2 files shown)


YijinL:
Ok

src/main/java/address/ui/Ui.java
```diff
@@ -0,0 +1,34 @@
...
+import javafx.stage.Stage;
+
+/**
+ * The UI of the app.
+ */
+public class Ui {
+    MainController mainController;
+
+    public Ui(MainApp mainApp, ModelManager modelManager, Config config){
+        BrowserManager.initializeJxBrowserEnvironment();
```
HansNewbie: This will fail. It needs to be in the original place because `init` of `MainApp` is special in a sense that it is not in UI thread.

This call needs to be in non-UI thread from the documentation, but it only works if called in `init` of `MainApp` (which is a non-UI thread). I tried to call this in a separate thread at about the same position as here and the application will get stuck because of deadlock (UI is initalizing browser but browser is waiting for UI to be initialized completely).

HansNewbie: Correction: this will fail *on Mac*. I will test to see if it still fails with this call.

damithc: I've already merged it. Let me know if it fails on mac.

HansNewbie: Works on Mac because this call is still in `init` method of `MainApp`. We may not be able to move it outside of `init`, though.

damithc: The code should move out of init method and buried somewhere in the browser component, but it should still be invoked from the init method. ui.init() --> browser.init() --> jxBrowswer init

HansNewbie: Noted. @yl-coder maybe take a look at it when you are refactoring browser.

yl-coder: this init method needs to be call in a non-ui thread. But i guess we can run it in a seperate thread in the browser init method.

damithc: The thread is decided by the calling method. You should be able to push it down to any class as long as the method chain starts with the init method of MainApp which is run on a non-UI thread.

yl-coder: If i'm not wrong, some resources can't be initialized in the mainApp init(). For example the Java WebView need to be initialize after the stage is created, which is the start() method. I believe @HansAdrian has tried many different location for this init method but can't seems to find a suitable place.

________________________________

HansNewbie: I only tried the initialization of jxBrowser, which only works in `MainApp` `init` method so what has been suggested actually is possible. Not sure about WebView as I was not involved.

yl-coder: Java WebView can't be initialized in non-ui thread. Another way will be to have two init method in the browserManager. one for calling in UI thread and one not. I guess this will complicate stuffs.
________________________________

PR #302, opened 2016-06-25
Refine updater component

damithc:
Some TODOs added. @HansNewbie to continue with the branch.
Areas to improve:
1. Error handling
2. Documentation (describe how the updater works)
3. Testing

Do we cater for these case?
1. a user misses a few versions due to prolonged disuse
2. a change in major version (not backward compatible)


src/main/java/address/updater/UpdateManager.java
```diff
@@ -107,6 +109,7 @@ private void checkForUpdate() {
         if (latestVersion.get().isEarlyAccess() != MainApp.IS_EARLY_ACCESS) {
             raise(new UpdaterFinishedEvent(MSG_DIFF_CHANNEL));
             logger.debug(MSG_DIFF_CHANNEL);
+            //TODO: how could this happen?
```
HansNewbie: This should never happen unless we mess up the update data. `assert false` seems to be too harsh as the chance that we might mess up the update data during release.

src/main/java/address/updater/UpdateManager.java
```diff
@@ -138,6 +141,7 @@ private void checkForUpdate() {
         raise(new UpdaterInProgressEvent("Downloading updates", 0.5));
         try {
             downloadAllFilesToBeUpdated(new File(UPDATE_DIR), filesToBeUpdated);
+            //TODO: is the progress bar updated after each file
```
HansNewbie: No. Right now the progress bar just stuck at 50% while downloading.

src/main/java/address/updater/model/LibraryDescriptor.java
```diff
@@ -15,6 +16,7 @@
     private OsDetector.Os os;
 
     public LibraryDescriptor() {}
+    //TODO: not used?
```
HansNewbie: Required for serialization.

src/main/java/address/updater/model/LibraryDescriptor.java
```diff
@@ -1,4 +1,5 @@
 package address.updater.model;
+//TODO: perhaps no need to have a model package here (to avoid confusion with other model packages)
```
HansNewbie: will change.

src/main/java/address/updater/JarUpdater.java
```diff
@@ -128,6 +129,7 @@ private void applyUpdateToAllFiles(String sourceDir, List<String> filesToBeUpdat
      * the process has not ended yet. As such, we will make several tries with wait.
      */
     private void applyUpdate(Path source, Path dest) throws IOException {
+        //TODO: this method is about moving a file rather than applying an update?
```
HansNewbie: Apply update is technically replacing the old JAR with a new JAR. This method has extra check if the JAR is already accessible (Windows specific problem where JAR in process cannot be modified). Will put comments.

src/main/java/installer/Installer.java
```diff
@@ -1,4 +1,5 @@
 package installer;
+//TODO: does this need to be a separate package?
```
HansNewbie: Technically this `Installer` has nothing to do with the application; it's purely for release/devops purpose. It's to create a self-extracting zip file. Hence I separated the package.

damithc: Can all that code be put in a separate updater package and take it out of addressbook? i.e. `hubturbo.updater`

src/main/java/address/updater/UpdateManager.java
damithc: We'll, logging a debug message (which won't even be enabled) is not the way to detect we've messed up the update :-p

src/main/java/address/updater/model/LibraryDescriptor.java
damithc: Add a comment.

HansNewbie:
Currently we cater for user misses a few versions. Not catering to change in major version (not backward compatible) yet.

src/main/java/address/updater/DependencyTracker.java
```diff
@@ -130,6 +129,7 @@ private void excludePlatformSpecificDependencies(List<String> dependencies) {
 
         try {
             updateData = JsonUtil.fromJsonString(json, UpdateData.class);
+            //TODO: is it possible to have a generic static method in StorageManager to read/write various json files?
```
HansNewbie: Should exception in read/write json files be handled in this static method or thrown back to caller?

src/main/java/address/updater/UpdateData.java
```diff
@@ -12,13 +12,14 @@
  */
 @JsonPropertyOrder({ "version", "mainApp", "libraries" })
 public class UpdateData {
+    //TODO: better class name?
```
HansNewbie: Can't think of better name. Is `VersionDescriptor` good enough? Since this will list version of application available and the download links of the files required for that version.

src/main/java/address/updater/DependencyTracker.java
damithc: Back to the caller

src/main/java/address/updater/UpdateData.java
damithc: Yes, that sounds more noun-like than the current one.

src/main/java/address/updater/DependencyTracker.java
HansNewbie: Should I replace all Json file handling in this PR or a separate PR? If separate PR, I will do replace all Json file handling in updater first.

src/main/java/installer/Installer.java
HansNewbie: Sure.

HansNewbie:
All TODOs have been addressed except documentation. Will write one now. Will do testing in a separate PR.

src/main/java/address/MainApp.java
```diff
@@ -73,45 +74,29 @@ protected void initComponents(Config config, UserPrefs userPrefs) {
...
-        alertMissingDependencies();
-        //TODO: should this be here? looks out of place
     }
 
     @Override
     public void start(Stage primaryStage) {
-        logger.info("Starting application: {}", Version.getCurrentVersion());
+        logger.info("Starting application: {}", MainApp.VERSION);
+        DependencyChecker dependencyChecker = new DependencyChecker(this::quit);
+        dependencyChecker.run();
```
damithc: Something like this?
```java
new DependencyChecker(REQUIRED_JAVA_VERSION, this::quit).verify();
```
Shouldn't this be in the init method?

src/main/java/address/MainApp.java
```diff
@@ -73,45 +74,29 @@ protected void initComponents(Config config, UserPrefs userPrefs) {
         syncManager = new SyncManager(config);
         keyBindingsManager = new KeyBindingsManager();
         updateManager = new UpdateManager();
```
damithc: UpdateManager should take the version as a parameter?

src/main/java/address/controller/StatusBarFooterController.java
```diff
@@ -136,6 +136,16 @@ public void handleUpdaterCompletedEvent(UpdaterFinishedEvent ufe) {
     }
 
     @Subscribe
+    public void handleUpdaterFailedEvent(UpdaterFailedEvent ufe) {
+        Platform.runLater(() -> {
+            updaterStatusBar.setText(ufe.toString());
+            updaterStatusBar.setProgress(0.0);
+            updaterStatusBar.setText("");
+            saveLocationLabel.setVisible(true);
```
damithc: What's this got to do with save location?

src/main/java/address/MainApp.java
HansNewbie: If it is in init method, I will need to use an alert dialog that is not from JavaFX but from Java Swing instead because at that point in that the UI-thread has not been created. Is that alright?

src/main/java/address/MainApp.java
HansNewbie: Will do.

src/main/java/address/controller/StatusBarFooterController.java
HansNewbie: I am following the original implementation of the `StatusBarFooterController`. I think the intention was not to show `saveLocationLabel` if updater is working, and only show it once updater is done. Hence the save location here.

src/main/java/address/MainApp.java
HansNewbie: I have moved it to init method and use Swing alert dialog.

src/main/java/address/controller/StatusBarFooterController.java
damithc: I see. @yl-coder in that case the name of the label should change.

src/main/java/address/updater/DependencyHistoryHandler.java
```diff
@@ -0,0 +1,118 @@
...
+            return;
+        }
+
+        Optional<List<String>> libraries = ManifestFileReader.getLibrariesInClasspathFromManifest();
+
+        if (!libraries.isPresent()) {
+            logger.debug("No libraries being used - should not happen");
+            assert false;
+            return;
+        }
```
damithc: This whole block can be reduced to this?
`assert libraries.isPresent() : "No libraries being used - should not happen";`

src/main/java/address/updater/DependencyHistoryHandler.java
```diff
@@ -0,0 +1,118 @@
...
+        } catch (IOException e) {
+            logger.debug("Failed to read dependencies from file");
+            e.printStackTrace();
+        }
+    }
+
+    public List<String> getCurrentVersionDependencies() {
+        return dependenciesForVersionsInUse.get(currentVersion);
+    }
+}
```
damithc: It feels like we should have a `Dependencies` object that could be put inside the `Version` object.

damithc: But ok if too hard to add.

src/main/java/address/updater/LocalUpdateSpecification.java
```diff
@@ -0,0 +1,27 @@
+package address.updater;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ *
+ */
```
damithc: ?

damithc:
No further comments from me

src/main/java/address/updater/DependencyHistoryHandler.java
HansNewbie: Noted.

src/main/java/address/updater/LocalUpdateSpecification.java
HansNewbie: Will add header comment.

src/main/java/address/updater/DependencyHistoryHandler.java
HansNewbie: I think putting in `Dependencies` object inside `Version` will unnecessarily complicate the `Version` object, since sometimes we use `Version` without any need about `Dependencies`. I will keep it this way for now.

docs/Updater.md
```diff
@@ -0,0 +1,19 @@
...
+<img src="images/How Updater Works.jpg" width="600">
+
+## Maintaining Backups and Dependencies
+A backup of the application will be created before applying update, that is, upon closing the application. Dependencies
+of backup versions are stored so that dependencies still used by backup versions will not be deleted erroneously.
+
+Several backups will be kept to allow user to roll back to older versions if update keeps failing. Once the amount of
+backups exceeds the limit of amount of backups stored (as specified in the code), the oldest backup version will be
+removed. If there are dependencies which are no longer used as the oldest backup version is removed, those dependencies
+will be removed as well so as not to take up disk space unnecessarily.
```
damithc: 1. use standard uml in the sequence diagram.
2. also describe a bit more about what are the names of different jar files, where the backups are stored, how are they named, how to revert to an older version, when does the updater runs, when is the update applied.

HansNewbie:
I have updated the docs.

damithc:
No more comments from me.

PR #303, opened 2016-06-25
Added a parent class UiController for controllers

damithc:
Trying to limit the coupling of EventManager to as fewer classes as possible


YijinL:
Are we still sticking to the singleton event manager?

damithc:
For the moment yes. It will be easier to change later if EventManager is referenced by fewer classes.

HansNewbie:
Looks good.

PR #307, opened 2016-06-26
Added a few GUI tests for key bindings

damithc:
Added some GUI tests for keybindings

### DIFF
--- a/src/main/java/address/storage/StorageManager.java
+++ b/src/main/java/address/storage/StorageManager.java
@@ -188,16 +188,20 @@ public class StorageManager extends ComponentManager {
     public void start() {
         logger.info("Starting storage manager.");
         loadDataFromFile(prefs.getSaveLocation());
-        raise(new SaveLocationChangedEvent(prefs.getSaveLocation()));
     }
 
     protected void loadDataFromFile(File dataFile) {
         try {
             logger.debug("Attempting to load data from file: {}", dataFile);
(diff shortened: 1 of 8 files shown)


src/main/java/address/storage/StorageManager.java
```diff
@@ -181,16 +181,20 @@ public static UserPrefs getUserPrefs(File prefsFile) {
     public void start() {
         logger.info("Starting storage manager.");
         loadDataFromFile(prefs.getSaveLocation());
-        raise(new SaveLocationChangedEvent(prefs.getSaveLocation()));
```
damithc: I removed this because this line doesn't fit here. i.e. there is no change to save location during startup.
If this cause regressions, it needs to be fixed without adding this back.

YijinL:
Looks good

src/main/java/address/storage/StorageManager.java
HansNewbie: This event call is to initialise the label at the footer on save file location. If this is removed, the label will be empty.

HansNewbie:
No more comments.

src/main/java/address/storage/StorageManager.java
damithc: Initializing the label should not be done using this event. We should not fire the event for the sole purpose of causing side effects. The label should be initialized simply by getting the value from model manager.

Issue #289, opened 2016-06-23
Add a full system test


damithc:
In the first round, we can simply start the full app -- as similar to user environment as possible -- and do some ui tasks (no need to verify anything).
We can add verifications later.

damithc:
@m133225 can do this one?

m133225:
Is this supposed to just randomly perform UI tasks?

damithc:
Yes, that would be the simplest first version.

damithc:
@m133225 you are doing this right?

m133225:
Yes, will be working on this.

PR #308, opened 2016-06-26
A test in KeyBindingsManagerApiTest fails locally #293

### LINKED ISSUE #293: A test in KeyBindingsManagerApiTest fails locally
The test `keyEventDetected_definedBinding_matchingEventRaised` fails locally on Mac.

damithc:
Fixes #293

### DIFF
--- a/src/test/java/address/guitests/GuiTestBase.java
+++ b/src/test/java/address/guitests/GuiTestBase.java
@@ -5,20 +5,15 @@ import address.events.EventManager;
 import address.keybindings.KeyBinding;
 import address.keybindings.KeySequence;
 import address.model.datatypes.ReadOnlyAddressBook;
-import address.util.OsDetector;
 import address.util.TestUtil;
 import javafx.scene.input.KeyCode;
 import javafx.scene.input.KeyCodeCombination;
-import javafx.scene.input.KeyCombination;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.testfx.api.FxRobot;
 import org.testfx.api.FxToolkit;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
 import java.util.concurrent.TimeoutException;
 
 public class GuiTestBase extends FxRobot {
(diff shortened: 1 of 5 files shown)


damithc:
@m133225 @HansNewbie see if this fix works in your Macs

m133225:
Seems to be working.

src/test/java/address/util/PlatformSpecific.java
```diff
@@ -0,0 +1,100 @@
+package address.util;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+
+
+public final class PlatformSpecific {
```
HansNewbie: I actually already created a class similar to this, named `OsDetector` in `util` of production since we have platform specific codes.

HansNewbie:
The `KeyBindingsManagerApiTest` no longer fails.

damithc:
OK, now that I know the fix works, I'll try to apply the technique in a tidier way.

damithc:
Waiting for #289 to finish first (and reuse the solution)

damithc:
@m133225 merge the keycombination fix as a separate PR so that this PR can use it?

m133225:
Done in #332.

damithc:
@m133225 @HansNewbie see if this passes on Mac.
I tried to reuse @m133225 's previous solution, but it didn't work for keybinding tests. Had to combine the two solutions in the end.

m133225:
The `KeyBindingManagerApiTest` itself works both in headless and headful, but I think there is still a problem with the shortcut when calling `push` in GUI tests.
- Seems like META + KeyCode responds correctly in headful, and COMMAND + KeyCode responds correctly in headless
  - With this PR's solution (which converts `SHORTCUT` to `META`),
    - Headless mode
      -`push(COMMAND, A)` passes
      -`push(KeyCodeCombination(A, SHORTCUT/META))` fails
    - Headful mode
      -`push(COMMAND, A)` fails
      - `push(KeyCodeCombination(A, SHORTCUT/META))` passes

damithc:
Should we use `COMMAND` instead? If that doesn't work, we'll have to select based on the test mode :-p

m133225:
Do you mean converting `SHORTCUT` to `COMMAND` (like the current master)? That still shouldn't work in headful mode.

HansNewbie:
Works on Mac for the `KeyBindingManagerApiTest`.

damithc:
In that case we'll probably need to vary the keys based on the test mode too. May be introduce a `TestEnv` class (similar to `OsDetector` and use it in the relevant methods? @m133225 can you continue this PR?

m133225:
Will we still manually fire events in `KeyBindingManagerApiTest` if it is fixed? If we are, we could merge this first (without the `push` changes) and work on the headful/headless issue in another PR.

m133225:
@damithc Updated it. Works locally on Mac in both headless and headful modes. Works on Travis in headless mode too. Could you try `TagEditDialogTest` (as well as `KeyBindingsManagerApiTest`) on Windows?

damithc:
Headless: passed in one try, both times I tried.
Headfull: tried three times. First two times some tests failed, but the failing tests passed when retried. Third time all tests passed in one try.
So I guess can merge.

PR #313, opened 2016-06-27
Move browser init method

yl-coder:
Moved the initialization of the jxBrowserEnvironment to a more appropriate place. Added a start method in the browserManager for initializing init tasks that require to be run in the UI-thread.


yl-coder:
Created wrongly. Discarding.

PR #314, opened 2016-06-27
Add a full system test #289

### LINKED ISSUE #289: Add a full system test

m133225:
Fixes #289

**Problems noted**
For *headless* tests:
1. Text fields have to be manually clicked for focus. Components requesting for focus do not seem to be working.
2. Pressing `ESC` when in the `TagListDialog` does not close the dialog.
3. Showing file chooser will throw an exception only on Mac. See `MonocleApplication::staticCommonDialogs_showFileChooser`. Shouldn't it also not work for Windows and Travis?


src/test/java/address/guitests/FullSystemTest.java
```diff
@@ -0,0 +1,49 @@
...
+        .clickOn("New").clickOn("#firstNameField").write("Ming").clickOn("OK").targetWindow("Invalid Fields").clickOn("OK")
+                .clickOn("#lastNameField").write("Lee").clickOn("OK")
+        .clickOn("File").clickOn("[Local] Save As...").sleep(1000).type(KeyCode.S).type(KeyCode.ENTER)
+        .clickOn("File").clickOn("[Local] Save").clickOn("File").clickOn("Exit");
+    }
+
+    @After
+    public void cleanup() {
+        File file = new File("src/test/data/sandbox/s.xml");
+        file.delete();
```
damithc: 1. There is a way to specify file name only. See other tests for example.
2. Perhaps no need to clean up the file? It's git ignored and will be overwritten next time. Sometimes it helps to have the temp files around after the test, for troubleshooting.

m133225: Regarding 2, the test will be incorrect when run the second time since it needs to go through an additional confirmation dialog to replace the previous file.

src/test/java/address/guitests/GuiTestBase.java
```diff
@@ -106,6 +107,12 @@ protected void pressKeyCombo(KeyBinding kb){
         return keys;
     }
 
+    public KeyCodeCombination shortcut(KeyCode keyCode) {
+        if (OsDetector.isOnMac()) {
+            return new KeyCodeCombination(keyCode, KeyCodeCombination.META_DOWN);
+        }
+        return new KeyCodeCombination(keyCode, KeyCodeCombination.CONTROL_DOWN);
+    }
```
damithc: Something like this?
```java
public FxRobot push(KeyCombination keys){
    return super.push(getPlatformSpecificKeyCombination(keys));
}

private KeyCombination getPlatformSpecificKeyCombination(KeyCombination keys){
    //If the combination has SHORTCUT and platform is Mac, replace with META
}
```

src/test/java/address/guitests/FullSystemTest.java
damithc: In that case, do the clean up at the beginning, not at the end. BTW, you can get GuiTestBase to set the file and the data to set to whatever you want.

src/test/java/address/guitests/GuiTestBase.java
m133225: But it seems unnecessarily long during the tests as we have to create new `KeyCodeCombination`s in the tests, when instead we can just call `shortcut(KeyCode)`.

m133225:
Couple of issues in headless mode:
- `TagSelectionEditDialog`
- `TagList`
- `shortcut(KeyCode)
- `Save As` dialog
Can't really debug without implementing screenshots.

src/test/java/address/guitests/GuiTestBase.java
damithc: As discussed, put this 'scrubbing' method inside the parent class.

m133225:
Waiting for #293 since a few things were migrated over there instead.

damithc:
Looks like it finally passed? hope it's not a fluke :-p

m133225:
This still fails when run in headless mode on Mac, due to the file chooser.

Trying to identify and document problems encountered during testing. Will update the opening comment.

damithc:
Seems stable on Travis. I ran it three times on travis and it passed all three times.

damithc:
Ideally, all tests should pass on all platforms in both modes. 
If that can't be done, an acceptable compromise is to disable some tests on some mode+platform configurations. It should be fine as long as all tests pass in travis. 
For example, we can disable a certain test on mac+headless.

m133225:
I've attempted to isolate problematic aspects and completely exclude them from the full system test. The tests should be able to run fine on all platforms for the moment.

Will introduce them back and disable them on some mode+platform if they cannot be solved soon.

src/test/java/address/guitests/FullSystemTest.java
```diff
@@ -0,0 +1,95 @@
...
+        .clickOn("Muster").type(KeyCode.E).sleep(1000)
+                .clickOn("#firstNameField").push(KeyCode.SHORTCUT, KeyCode.A).eraseText(1).write("John")
+                .clickOn("#lastNameField").eraseText(6).write("Tan")
+                .clickOn("#cityField").write("Singapore")
+                .clickOn("#githubUserNameField").write("john123");
+        verifyThat("#firstNameField", hasText("John"));
+        verifyThat("#lastNameField", hasText("Tan"));
+        verifyThat("#cityField", hasText("Singapore"));
+        verifyThat("#githubUserNameField", hasText("john123"));
+                clickOn("#tagList")
```
damithc: Why is this indented?

damithc: More generally, are there any additional indentation styles we want to follow in gui tests?

damithc:
No more comments for this version

src/test/java/address/guitests/FullSystemTest.java
m133225: Indented to show that we are still performing the operation in the `PersonEditDialog` (like editing the other fields). Not too sure what's a good style for this since we have to cut off the chain at some points to verify behaviour (if not it is really difficult to debug).

damithc: May be not indent as it is the start of a new statement.  We can use blank lines and comments to demarcate code blocks that are doing related things.

PR #315, opened 2016-06-27
Moved init method into browserManager.

yl-coder:
Moved init method into browserManager. Created a start method to contain init tasks that require to be run in UI-thread.

### DIFF
--- a/src/main/java/address/browser/BrowserManager.java
+++ b/src/main/java/address/browser/BrowserManager.java
@@ -29,6 +29,7 @@ import java.util.stream.Collectors;
 
 /**
  * Manages the AddressBook browser.
+ * To begin using this class: call start() once.
  */
 public class BrowserManager {
 
@@ -58,7 +59,7 @@ public class BrowserManager {
     };
 
     public BrowserManager(ObservableList<ReadOnlyViewablePerson> filteredPersons) {
-        initializeBrowser();
+        initializeJxBrowserEnvironment();
         this.selectedPersonUsername = new SimpleStringProperty();
         this.filteredPersons = filteredPersons;
     }
@@ -67,7 +68,7 @@ public class BrowserManager {
      * Initialize the application to use jxBrowser.
      * This must be called in a non-ui thread.
      */
-    public static void initializeJxBrowserEnvironment(){
+    private static void initializeJxBrowserEnvironment(){
         if (Environment.isMac()) {
             BrowserCore.initialize();
         }
@@ -75,7 +76,10 @@ public class BrowserManager {
         LoggerProvider.setLevel(Level.SEVERE);
     }
 
-    public void initializeBrowser() {
+    /**
+     * Starts the browser manager.
+     */
+    public void start() {
         String headlessProperty = System.getProperty("testfx.headless");
         if (headlessProperty != null && headlessProperty.equals("true")) {
             logger.info("Headless mode detected, not initializing HyperBrowser.");
--- a/src/main/java/address/browser/BrowserManager.java
+++ b/src/main/java/address/browser/BrowserManager.java
@@ -59,7 +59,6 @@ public class BrowserManager {
     };
 
     public BrowserManager(ObservableList<ReadOnlyViewablePerson> filteredPersons) {
-        initializeJxBrowserEnvironment();
         this.selectedPersonUsername = new SimpleStringProperty();
         this.filteredPersons = filteredPersons;
     }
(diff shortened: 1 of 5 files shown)


yl-coder:
@HansNewbie Can help to review this? thanks.

src/main/java/address/controller/MainController.java
```diff
@@ -82,7 +82,7 @@ public MainController(MainApp mainApp, ModelManager modelManager, Config config)
     public void start(Stage primaryStage) {
         logger.info("Starting main controller.");
         this.primaryStage = primaryStage;
-        this.browserManager.initializeBrowser();
+        this.browserManager.start();
```
damithc: Shouldn't this be `init()`?

yl-coder:
There is some issue, hold on.

HansNewbie:
I think you will need `BrowserManager.init()` where inside you will call `initJxBrowserEnv()`. Then call that `init` in `MainApp.init()`. I guess can keep the `start` to be consistent with `MainController.start`.

src/main/java/address/browser/BrowserManager.java
```diff
@@ -58,7 +59,7 @@
     };
 
     public BrowserManager(ObservableList<ReadOnlyViewablePerson> filteredPersons) {
-        initializeBrowser();
+        initializeJxBrowserEnvironment();
```
damithc: Shouldn't this be conditional? i.e. only if the jxbrowser is being used

damithc:
Added some prelim comments.

src/main/java/address/controller/MainController.java
yl-coder: I will have two method:
initBrowser() : to run init tasks in non-ui thread
start(): to run init tasks in ui thread

The start method in the MainController is in the ui thread.

src/main/java/address/browser/BrowserManager.java
yl-coder: Calling BrowserManager.init() before the browserManager instance is quite odd. I'm calling the init after the browserManager instance is created at mainController.

Yes, I will add in the conditional.

src/main/java/address/browser/BrowserManager.java
```diff
@@ -67,15 +67,18 @@ public BrowserManager(ObservableList<ReadOnlyViewablePerson> filteredPersons) {
      * Initialize the application to use jxBrowser.
      * This must be called in a non-ui thread.
      */
-    public static void initializeJxBrowserEnvironment(){
+    public static void initBrowser(){
```
HansNewbie: Update comments.

HansNewbie:
Tell me if it is ready. I will test on Mac.

src/main/java/address/browser/BrowserManager.java
```diff
@@ -57,25 +60,30 @@
...
     }
 
     /**
      * Initialize the application to use jxBrowser.
      * This must be called in a non-ui thread.
      */
-    public static void initializeJxBrowserEnvironment(){
-        if (Environment.isMac()) {
-            BrowserCore.initialize();
+    public void initBrowser(){
```
HansNewbie: Don't forget to update the header comment.

yl-coder:
@HansNewbie Can help to try if the test case runs on ur mac

HansNewbie:
Pass with flying colours.

yl-coder:
@damithc Anymore comments before i merged this PR?

src/main/java/address/browser/BrowserManager.java
```diff
@@ -58,32 +61,36 @@
...
-        if (Environment.isMac()) {
-            BrowserCore.initialize();
+    public static void initBrowser(){
+        if (type == BrowserType.FULL_FEATURE_BROWSER) {
+            if (Environment.isMac()) {
+                BrowserCore.initialize();
+            }
+            logger.debug("Suppressing browser logs");
+            LoggerProvider.setLevel(Level.SEVERE);
         }
```
damithc: This is better, but even better if you use a polymorphic call.
i.e. init() method is implemented in both types of browsers and you call the init() method here without knowing which type of a browser it is.

yl-coder: Do you mean to replace it with embeddedBrowser.init() here and implement the init method in respective concrete class of embeddedBrowser? The embeddedbrowser instance will only be created at the start() method. In which, the UI-thread will be started. And this block of codes need to run on a non-UI thread.

damithc: Yes, something like that. Never mind, we'll figure that out later.

damithc:
No other comments from me.

PR #320, opened 2016-06-27
Disabled TOTD feature temporarily

damithc:
I don't think we have time to polish this enough to include in V1.0
We'll add it back in a later version


HansNewbie:
Looks good.

PR #327, opened 2016-06-28
Fix autoscroll problem when page has already been loaded before

yl-coder:
Added a new functionality for browser/page to be able to detect if a page is attached to a scene.

### DIFF
--- a/src/main/java/address/browser/BrowserManager.java
+++ b/src/main/java/address/browser/BrowserManager.java
@@ -108,7 +108,10 @@ public class BrowserManager {
         try {
             Page page = hyperBrowser.get().loadUrls(person.profilePageUrl(), listOfFutureUrl);
             GithubProfilePage gPage = new GithubProfilePage(page);
-            gPage.setPageLoadFinishListener(b -> Platform.runLater(() -> gPage.executePageLoadedTasks()));
+            if (!gPage.wasAutoScrollingSetup()){
+                gPage.setupAutoScrolling();
+            }
+            //gPage.setPageLoadFinishListener(b -> Platform.runLater(() -> gPage.executePageLoadedTasks()));
 
         } catch (IllegalArgumentException e) {
             e.printStackTrace();
--- a/src/main/java/address/browser/BrowserManager.java
+++ b/src/main/java/address/browser/BrowserManager.java
@@ -14,7 +14,6 @@ import com.teamdev.jxbrowser.chromium.BrowserCore;
 import com.teamdev.jxbrowser.chromium.LoggerProvider;
 import com.teamdev.jxbrowser.chromium.internal.Environment;
 
-import javafx.application.Platform;
 import javafx.beans.property.SimpleStringProperty;
 import javafx.beans.property.StringProperty;
 import javafx.beans.value.ChangeListener;
@@ -108,8 +107,8 @@ public class BrowserManager {
         try {
             Page page = hyperBrowser.get().loadUrls(person.profilePageUrl(), listOfFutureUrl);
             GithubProfilePage gPage = new GithubProfilePage(page);
-            if (!gPage.wasAutoScrollingSetup()){
-                gPage.setupAutoScrolling();
+            if (!gPage.wasPageAutomationSetup()){
+                gPage.setupPageAutomation();
             }
             //gPage.setPageLoadFinishListener(b -> Platform.runLater(() -> gPage.executePageLoadedTasks()));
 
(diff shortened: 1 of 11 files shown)


src/main/java/hubturbo/embeddedbrowser/jxbrowser/JxBrowserAdapter.java
```diff
@@ -79,12 +90,18 @@ public void executeCommand(int command) {
     }
 
     @Override
-    public void addLoadListener(EbLoadListener listener) {
-        this.listener = listener;
+    public void setLoadListener(EbLoadListener listener) {
+        this.loadListener = Optional.ofNullable(listener);
+        this.browserView.getBrowser().removeLoadListener(this);
```
HansNewbie: Why do we remove then add listener?

src/main/java/address/browser/BrowserManager.java
```diff
@@ -108,7 +107,9 @@ public synchronized void loadProfilePage(ReadOnlyViewablePerson person) {
         try {
             Page page = hyperBrowser.get().loadUrls(person.profilePageUrl(), listOfFutureUrl);
             GithubProfilePage gPage = new GithubProfilePage(page);
-            gPage.setPageLoadFinishListener(b -> Platform.runLater(() -> gPage.executePageLoadedTasks()));
+            if (!gPage.wasPageAutomationSetup()){
+                gPage.setupPageAutomation();
+            }
```
HansNewbie: checking if it's already setup should be inside `setup` call.

src/main/java/hubturbo/embeddedbrowser/jxbrowser/JxBrowserAdapter.java
yl-coder: setLoadListener may get called multiple times. I'm afraid that it adding the same instance twice may result in calling it twice. The documentation didn't state the details. 

I will remove the removeLoadListener and observe if it will get called multiple times if the listener is added multiple times.

yl-coder: Apparently, it is not. I will remove the removeLoadListener() line

yl-coder:
@HansNewbie Can help to review the new changes. The major part is to clean up the browser listener threads when it is going to be pushed back to the inActive pool of browser.

src/main/java/hubturbo/embeddedbrowser/page/Page.java
```diff
@@ -107,6 +108,11 @@ public boolean verifyPresenceByIds(String id){
 
     @Override
     public void setPageLoadFinishListener(EbLoadListener listener){
-        this.browser.addLoadListener(listener);
+        this.browser.setLoadListener(listener);
+    }
+
+    @Override
+    public void setPageAttachedToSceneListener(EbAttachListener listener) {
+        browser.setAttachListener(listener);
```
HansNewbie: to be consistent, `this.browser`.

src/main/java/address/browser/page/GithubProfilePage.java
```diff
@@ -31,11 +35,31 @@ public boolean isValidGithubProfilePage(){
...
+            this.setPageLoadFinishListener(e -> Platform.runLater(this::executePageLoadedTasks));
+            this.setPageAttachedToSceneListener(() -> Platform.runLater(this::executePageLoadedTasks));
+            this.wasAutoScrollingSetup = true;
+            //If page has already been loaded at the point of setting up page automation.
+            //Trigger initial automation.
+            if (!page.isPageLoading()) {
+                this.executePageLoadedTasks();
+            }
+        }
+    }
```
HansNewbie: no more lock?

src/main/java/address/browser/BrowserManager.java
```diff
@@ -47,7 +46,9 @@
         try {
             URL url = new URL(GITHUB_ROOT_URL + newValue);
             if (!UrlUtil.compareBaseUrls(hyperBrowser.get().getDisplayedUrl(), url)) {
-                hyperBrowser.get().loadUrl(url);
+                Page page = hyperBrowser.get().loadUrl(url);
+                GithubProfilePage gPage = new GithubProfilePage(page);
+                gPage.setupPageAutomation();
```
HansNewbie: Do we need this setup here? Since this is also called below.

src/main/java/address/browser/page/GithubProfilePage.java
yl-coder: I think we can introduce the lock when we need it. Currently we won't have any concurrency issues now

HansNewbie:
No more comments.

HansNewbie:
Don't forget to squash the commits.

PR #329, opened 2016-06-28
Fix bug where nearest hour fails at 11pm

HansNewbie:
The app cannot run because at 11pm the nearest hour creation fails since it will be at hour 24 while the limit is 0 - 23.

### DIFF
--- a/src/main/java/address/sync/cloud/CloudRateLimitStatus.java
+++ b/src/main/java/address/sync/cloud/CloudRateLimitStatus.java
@@ -48,8 +48,9 @@ public class CloudRateLimitStatus {
 
     private long getNextResetTime() {
         LocalDateTime curTime = LocalDateTime.now();
+        LocalDateTime nextHour = curTime.plusHours(1);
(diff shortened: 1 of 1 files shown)


HansNewbie:
@m133225 can help review? Not sure if my correction will destroy stuffs.

m133225:
Hmm, doesn't it automatically resolve to the next day's 0h?

src/main/java/address/sync/cloud/CloudRateLimitStatus.java
```diff
@@ -48,8 +48,9 @@ private ZoneOffset getSystemTimezone() {
 
     private long getNextResetTime() {
         LocalDateTime curTime = LocalDateTime.now();
+        LocalDateTime nextHour = curTime.plusHours(1);
```
m133225: Think can just call `LocalDateTime.now().plusHours(1)`

HansNewbie: Sure.

PR #341, opened 2016-06-29
Initialize save location in the footer status bar #311

### LINKED ISSUE #311: Initialize save location in the footer status bar
The initialization got removed in a #307

yl-coder:
Fixes #311

### DIFF
--- a/src/main/java/address/controller/MainController.java
+++ b/src/main/java/address/controller/MainController.java
@@ -113,8 +113,6 @@ public class MainController extends UiController{
             // Load root layout from fxml file.
             FXMLLoader loader = loadFxml(fxmlResourcePath);
             rootLayout = loader.load();
-            SplitPane pane = (SplitPane) rootLayout.lookup("#splitPane");
-            pane.setDividerPositions(0.3f);
 
             // Show the scene containing the root layout.
             Scene scene = new Scene(rootLayout);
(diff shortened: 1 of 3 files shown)


src/main/java/address/controller/MainController.java
```diff
@@ -185,7 +183,7 @@ private void showFooterStatusBar() {
             GridPane gridPane = loader.load();
             gridPane.getStyleClass().add("grid-pane");
             StatusBarFooterController controller = loader.getController();
-            controller.initStatusBar(config);
+            controller.initStatusBar(config, modelManager.getPrefs());
```
HansNewbie: Should we name the method as `setConnections`? To be consistent with other controllers. If there is other initialization needed, maybe have `start` after `setConnections` or put in constructor.

yl-coder: changed to init()

src/main/java/address/controller/StatusBarFooterController.java
```diff
@@ -57,27 +57,28 @@ private void syncSchedulerOnTimeOut() {
...
-    public void initStatusBar(Config config) {
-        updateWithConfigValues(config);
+    public void init(long updateInterval, String saveLocation) {
+        initSyncTimer(updateInterval);
+        initStatusBar();
+        initSaveLocationLabel(saveLocation);
+    }
+
+    private void initSaveLocationLabel(String saveLocation) {
+        secondaryStatusBarLabel.setText(SAVE_LOC_TEXT_PREFIX + saveLocation);
```
HansNewbie: can call `updateSaveLocationDisplayHere`? So the format will always be the same.

HansNewbie:
No more comments

PR #343, opened 2016-06-30
Restructured test package structure to match production structure

YijinL:
GUI tests remain in their own packages.

Unit tests go from `address.unittests.*` to `address.*`. This allows easier testing and mocking of dependent classes in the same package.

### DIFF
--- a/src/test/java/address/JsonUtilTest.java
+++ b/src/test/java/address/JsonUtilTest.java
@@ -1,4 +1,4 @@
-package address.unittests;
+package address;
 
 import address.model.datatypes.AddressBook;
 import address.model.datatypes.ReadOnlyAddressBook;
--- a/src/test/java/address/PersonEditDialogModelTest.java
+++ b/src/test/java/address/PersonEditDialogModelTest.java
@@ -1,4 +1,4 @@
-package address.unittests;
+package address;
 
 import address.model.datatypes.tag.Tag;
 import address.model.TagSelectionEditDialogModel;
--- a/src/test/java/address/browser/BrowserManagerTest.java
+++ b/src/test/java/address/browser/BrowserManagerTest.java
@@ -3,11 +3,15 @@ package address.browser;
 import address.util.JavafxRuntimeRule;
 import hubturbo.embeddedbrowser.BrowserType;
 import javafx.collections.FXCollections;
+import org.junit.BeforeClass;
 import org.junit.Rule;
 import org.junit.Test;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 
 import static org.junit.Assert.assertNotNull;
 
@@ -23,6 +27,19 @@ public class BrowserManagerTest {
      */
     public JavafxRuntimeRule javafxRule = new JavafxRuntimeRule();
 
+    @BeforeClass
+    public void setup(){
+        ExecutorService executor = Executors.newFixedThreadPool(1);
+
+        Runnable task = () -> new BrowserManager(FXCollections.emptyObservableList(),
+                1, BrowserType.FULL_FEATURE_BROWSER).initBrowser();
+        Future<?> future = executor.submit(task);
+        try {
+            future.get();
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+    }
 
     @Test
     public void testNecessaryBrowserResources_resourcesNotNull() throws NoSuchMethodException,
--- a/src/test/java/address/browser/BrowserManagerTest.java
+++ b/src/test/java/address/browser/BrowserManagerTest.java
@@ -9,9 +9,6 @@ import org.junit.Test;
 
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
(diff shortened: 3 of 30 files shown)


YijinL:
@HansNewbie `./gradlew clean headless` works on my windows machine, can you pull the branch and see if it works on mac?

YijinL:
@m133225 All I did was move some folders around in the test directory (you can see the changelog), but it seems to break stuff. Can you take a look at both travis build logs, see if you know what's happening?

m133225:
Can't figure out why. Could be just that the current assigned travis worker is unable to see the last card (`Cornelia` in `PersonEditDialogTest` is the 4th card).

damithc:
Some non-determinism and state-leaks may be creeping into tests?. These only start acting up under certain conditions, like when the environment is slower than usual.

damithc:
Sync with master and try again?

damithc:
Perhaps the renaming of packages causes a change in the order of tests, bringing into play a state leak or timing issue that was hidden before?

YijinL:
Maybe, but it's strange that the travis push branch consistently fails and the pull branch consistently passes

damithc:
The two branches must be using two different test orderings consistently?

damithc:
@YijinL @m133225 @yl-coder 
I think I narrowed down the problem by a process of trial-and-error. Realizing that the order of test could be the cause, and the tests are run based on the alphabetical order of packages, I kept renaming packages to change the order of testing. I found that placing the 'browser' package before the 'gui*' packages causes the tests to fail. Out of the 3 classes in the `browser` package, `BrowserManagerTest` and the `EmbeddedBroswerObjectMapperTest` are the problematic ones. Running either one before the gui tests will cause some of the gui tests to fail.

YijinL:
so the pr build and push build use different ordering?

yl-coder:
Could be the browser instance taking too much memory and resources. Or In browsermanagertest, it runs a rule that creates a fx application theads manually so that I could create fx view and somehow this interfere with the guitest.

Get Outlook for Android<https://aka.ms/ghei36>



On Sat, Jul 2, 2016 at 12:28 PM +0800, "Damith C. Rajapakse" <notifications@github.com<mailto:notifications@github.com>> wrote:


@YijinL<https://github.com/YijinL> @m133225<https://github.com/m133225> @yl-coder<https://github.com/yl-coder>
I think I narrowed down the problem by a process of trial-and-error. Realizing that the order of test could be the cause, and the tests are run based on the alphabetical order of packages, I kept renaming packages to change the order of testing. I found that placing the 'browser' package before the 'gui*' packages causes the tests to fail. Out of the 3 classes in the browser package, BrowserManagerTest and the EmbeddedBroswerObjectMapperTest are the problematic ones. Running either one before the gui tests will cause some of the gui tests to fail.

-

damithc:
@YijinL what do you mean pr build and push build?
Yes, only one of the two builds fail. I assume one was using the headless mode and the other is using the headful mode. @m133225 how do the two builds differ?

m133225:
They don't differ in anything (one builds when we push onto a branch in the main repo, the other one when we create a PR). Technically we should be disabling the push one for all branches except for master.

damithc:
I see. In that case it is indeed strange that the PR build passes consistently but the push build fails consistently. There must be a hidden difference between the two.

damithc:
I tried diffing the log files from the pr build and push build. It seems they are run on slightly different code bases. One uses the branch while the other does a temp merge with master first. Not sure how it affects the test result though.

YijinL:
back when the build was run I had rebased on master. will run after rebasing again

damithc:
Confirmed, the two builds use different test orderings (found by comparing the two log files). The push-build goes alphabetically while the pr-build uses some other ordering in which gui tests are run first.

damithc:
I think the ideal solution is to fix the state leak between browser tests and gui tests. Failing that, we need to ensure gui tests are run before browser tests.

yl-coder:
The browser tests will initialize the javafx runtime as some of the test methods require to be ran in the FX UI thread. I guess this somehow conflict with the guitest, as it will initialize the javafx runtime again.

YijinL:
Likewise my model tests will do the same. Because the code we test calls
stuff like platform.runLater...

damithc:
1. Try to run those tests before the gui tests locally (using gradle) and see if you can reproduce the problem.
2. May be shutdown the javafx env after your tests?

damithc:
I managed to reproduce the problem locally. Steps:
Add a test suite (for specifying the test order):
```java
package address;

import address.guitests.FullSystemTest;
import address.guitests.PersonOverviewTest;
import address.guitests.TagEditDialogTest;
import address.unittests.browser.BrowserManagerTest;
import org.junit.runner.RunWith;
import org.junit.runners.Suite;

@RunWith(Suite.class)
@Suite.SuiteClasses({
        BrowserManagerTest.class,
        PersonOverviewTest.class
})

public class OrderedTests {
    // the class remains empty,
    // used only as a holder for the above annotations
}
```

Run the test suite in gradle `gradlew clean headless --tests address.OrderedTests`

yl-coder:
@damithc I managed to solved the issue. By using TestFx framework method to initialize the runtime.  The changes could be found in JavafxRuntimeRule.java 

Any comments before we merged this?

damithc:
Looks like the problem is fixed now. Well done!

m133225:
I think a few (new?) tests were missed out when moving the tests. e.g. Parser unit test. Will handle this in #394.

PR #345, opened 2016-06-30
Refactoring browser component

yl-coder:
To improve browser testing by passing in browserConfig to the browserManager constructor.


yl-coder:
I'm not sure having a browserManagerUtil is a good idea? The idea is to improve testability without using Reflection

yl-coder:
Open for comments to refactor other areas of the browser component

damithc:
Add values to the main Config class rather than create a new config class just for the browser? Later on we can use interfaces to ensure the Browser component sees only the values it needs.

yl-coder:
Ok sure. Removed BrowserConfig

src/main/java/address/controller/MainController.java
```diff
@@ -76,7 +76,8 @@ public MainController(MainApp mainApp, ModelManager modelManager, Config config)
         this.modelManager = modelManager;
         this.config = config;
         this.personList = modelManager.getAllViewablePersonsReadOnly();
-        this.browserManager = new BrowserManager(personList);
+        this.browserManager = new BrowserManager(personList,
+                                                 config.getBrowserNoOfPages());
```
damithc: Doesn't look like you need a line break here.

src/main/java/address/util/Config.java
```diff
@@ -25,6 +26,7 @@
     public Level currentLogLevel = DEFAULT_LOGGING_LEVEL;
     public HashMap<String, Level> specialLogLevels = DEFAULT_SPECIAL_LOG_LEVELS;
     private File prefsFileLocation = new File("preferences.json"); //Default user preferences file
+    public int browserNoOfPages = DEFAULT_BROWSER_NO_OF_PAGES;
```
damithc: You can also add browser type (webview, jxbrowser) as a config parameter.

src/main/java/address/util/UrlUtil.java
```diff
@@ -19,4 +20,33 @@ public static boolean compareBaseUrls(URL url1, URL url2) {
...
+    /**
+     * Gets a list of URL that may be potentially be used in the future.
+     * The {@link URL} that will be used at the current moment must be in the param urls. And it will be to determine
+     * which URL may be used in the future.
+     * @param urls The list of URL that contains URL that may be potentially used in the future.
+     * @param selectedUrl The index of the current URL in the list(urls) that will be used at this moment.
+     * @param noOfFuturisticUrl The number of future URLs you wish to obtain from the list.
+     * @return A list of url that may be potentially used in the future. Range of list size: [0, noOfFuturisticUrl]
+     */
+    public static List<URL> getFuturisticUrls(List<URL> urls, int selectedUrl, int noOfFuturisticUrl) {
```
damithc: I don't think `futuristic` is the right term here.

damithc:
No more comments for this version.

src/main/java/address/controller/MainController.java
yl-coder: Ok, fixed

src/main/java/address/util/UrlUtil.java
yl-coder: ok changed to future

src/main/java/address/util/Config.java
yl-coder: Noted. Added

src/main/java/address/util/UrlUtil.java
```diff
@@ -16,7 +17,36 @@ public static boolean compareBaseUrls(URL url1, URL url2) {
         boolean isSameBaseUrl = url1.getHost().toLowerCase().replaceFirst("www.", "")
                                 .equals(url2.getHost().replaceFirst("www.", "").toLowerCase())
                                 && url1.getPath().replaceAll("/", "").toLowerCase()
-                                   .equals(url2.getPath().replaceAll("/", "").toLowerCase());
+                                .equals(url2.getPath().replaceAll("/", "").toLowerCase());
```
damithc: URL class might have methods that could make this comparison simpler. Try googling for a better solution.

PR #349, opened 2016-06-30
Add tests to StorageManager

HansNewbie:


HansNewbie:
Added more tests to `StorageManager`. Some methods are not tested  because of not being able to mock `static void` methods.

src/test/java/address/unittests/storage/StorageManagerTest.java
```diff
@@ -20,25 +24,36 @@
...
 @PrepareForTest(XmlFileStorage.class)
 public class StorageManagerTest {
 
     private static final File DUMMY_DATA_FILE = new File(TestUtil.appendToSandboxPath("dummyAddressBook.xml"));
     private static final File DUMMY_PREFS_FILE = new File(TestUtil.appendToSandboxPath("dummyUserPrefs.json"));
+    private static final File SERIALIZATION_FILE = new File(TestUtil.appendToSandboxPath("serialize.json"));
+    private static final File INEXISTENT_FILE = new File(TestUtil.appendToSandboxPath("inexistent"));
+    private static final File NON_JSON_FILE = new File(TestUtil.appendToSandboxPath("non.json"));
+
+
```
yl-coder: extra line

yl-coder:
No further comments

src/test/java/address/util/TestSerializationClass.java
```diff
@@ -0,0 +1,73 @@
...
+
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+
+/**
+ * A class used to test serialization and deserialization
+ */
+public class TestSerializationClass {
```
damithc: Let's try to stick to 'one unit test class per each functional class' and name the test class to match the functional class.

src/test/java/address/unittests/storage/StorageManagerTest.java
```diff
@@ -62,6 +76,20 @@ public void setup(){
...
+        PowerMockito.verifyPrivate(storageManagerSpy).invoke("loadDataFromFile", userPrefsMock.getSaveLocation());
+    }
+
+    @Test
+    public void saveAddressBook_noException() throws IOException, DataConversionException {
+        StorageManager.saveAddressBook(DUMMY_DATA_FILE, EMPTY_ADDRESSBOOK);
+
+        PowerMockito.verifyStatic();
+        FileUtil.createIfMissing(DUMMY_DATA_FILE);
+    }
```
damithc: If a method is not tested, create an empty test method and explain why it is not tested as a comment inside it.

src/test/java/address/util/TestSerializationClass.java
HansNewbie: This  class is for testing purpose only. Basically this class is used to test serialization and deserialization so that we don't need to use our models for testing the serializers. For example, if we port the serializer to another app, we don't need to rewrite the tests to replace `Addressbook` with another class because the tests depend on this class instead of `Addressbook`.

Of course, we need to write another test to ensure `Addressbook` can be serialized properly, too.

damithc: I see. Probably the class name is misleading. When you start the name with 'Test' it sounds like a verb.

HansNewbie: Is `SerializationClass` a better name?

HansNewbie: Will rename to `SerializableTestClass`.

PR #359, opened 2016-07-01
Fix regression caused by PR #341

yl-coder:


yl-coder:
@HansNewbie Can help to review this. Didn't handle the case when File is null during PR #341 therefore causing the regression.

HansNewbie:
Looks good.

PR #366, opened 2016-07-01
Gui tests on Windows: Only partial GUI is visible #333

### LINKED ISSUE #333: Gui tests on Windows: Only partial GUI is visible
Browser and the title bar is missing

yl-coder:
Fixes #333


damithc:
Can merge. In a separate PR, we can set the initial size to a more suitable size. The current one is too small.

PR #369, opened 2016-07-01
Connect new remote api to model #340

### LINKED ISSUE #340: Communication between remote and model
#### Sync down (auto)
- An event is raised by REMOTE and captured by MODEL
- The event contains the changes in the remote since the last request

**Problems for discussion**

1. Last-synced timestamp should be stored in MODEL (need to save to disk) so the sync-down scheduling thread must be able to retrieve the last-synced time from MODEL. My suggestion is for the REMOTE component's constructor to take in a `Supplier<LocalDateTime>` which will be called by the sync-thread every cycle.
- Do we keep a single last-synced time for the entire addressbook/model? Or a last synced time for every data item?
- Need to decide on data format; eg. will remote also tell model what has been deleted?

#### Sync-up (user commands)
There are two types of remote requests here:

1. Right before sending the real change request to the remote, we will query the remote to check whether there has been a change since the last-synced timing before the user starts the edit/add/del.
- The real change request to the remote:
    - add a person
    - delete one person
    - edit one persons

**Proposal:**

- MODEL was raise an event containing a `CompletableFuture`. Once REMOTE captures the event, it will `CompletableFuture.complete` with the result, and MODEL will be informed since it was waiting using `Future.get`. Any changes to propose?
- We need to decide on the visible response format. Here is the current idea: both request types will involve a `CompletableFuture<Optional<ReadOnlyPerson>>`. MODEL will get the `Optional<ReadOnlyPerson>` from the future. 
    - An empty optional means there were no changes (request type 1), or the request has failed (request type 2).
    - A filled optional contains the updated person data (request type 1), or the newly created person on the remote (ADD), the deleted person (DEL), or the person after the successful edit (EDIT)

YijinL:
resolves #340

### DIFF
--- a/src/main/java/address/controller/RootLayoutController.java
+++ b/src/main/java/address/controller/RootLayoutController.java
@@ -122,7 +122,7 @@ public class RootLayoutController extends UiController{
 
         modelManager.setPrefsSaveLocation(file.getPath());
         // TODO temp patch to create cloud file if non found
-        raise(new CreateAddressBookOnRemoteRequestEvent(new CompletableFuture<Boolean>(), file.getPath()));
+        raise(new CreateAddressBookOnRemoteRequestEvent(new CompletableFuture<>(), file.getName()));
         raise(new SaveDataRequestEvent(file, modelManager));
     }
 
--- a/src/main/java/address/controller/RootLayoutController.java
+++ b/src/main/java/address/controller/RootLayoutController.java
@@ -1,6 +1,7 @@
 package address.controller;
 
 import address.MainApp;
+import address.events.CreateAddressBookOnRemoteRequestEvent;
 import address.events.LoadDataRequestEvent;
 import address.events.SaveDataRequestEvent;
 import address.keybindings.KeyBindingsManager;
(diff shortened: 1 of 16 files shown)


src/main/java/address/controller/RootLayoutController.java
```diff
@@ -119,6 +121,8 @@ private void handleSaveAs() {
         }
 
         modelManager.setPrefsSaveLocation(file.getPath());
+        // TODO temp patch to create cloud file if non found
+        raise(new CreateAddressBookOnRemoteRequestEvent(new CompletableFuture<>(), file.getName()));
```
damithc: I assume in the permanent version controllers will not deal with `CompletableFuture`s?

YijinL: Yes, this is a hack to let the user have some way of creating remote files when they never had one.

damithc:
No more comments from me.

YijinL:
@m133225 sync ups from user commands work fine, but the automatic syncdowns throw nullpointerexceptions. After merging this can take a look.

src/main/java/address/controller/RootLayoutController.java
damithc: I though this should be done automatically by the cloud simulator? i.e. create the file if it doesn't exist

m133225: No, it does not, just like how GitHub doesn't automatically create a repo if we try to push to a non-existing one.

PR #376, opened 2016-07-01
Gui tests with retry clicking

damithc:
I added a retry loop to the clickOn method in GUI tests, just to see if it improves the success rate.
Also eliminated the unittests package (to compare success rate with the other similar PR)


damithc:
Closing. No longer relevant.

PR #390, opened 2016-07-02
added scaffolding for reading values from the GUI

damithc:
Added some scaffolding to read values from the GUI. e.g. the name of the selected person.

### DIFF
--- a/src/main/java/address/MainApp.java
+++ b/src/main/java/address/MainApp.java
@@ -71,10 +71,14 @@ public class MainApp extends Application {
 
         modelManager = new ModelManager(userPrefs);
         storageManager = new StorageManager(modelManager::resetData, config, userPrefs);
-        ui = new Ui(this, modelManager, config);
+        ui = createUi(this, modelManager, config);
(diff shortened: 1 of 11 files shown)


damithc:
@m133225 do you think this kind of scaffolding is needed or we can do this stuff using testFx alone? 
I'll add the comments if we decide to keep this code.
@YijinL this PR touches the model too, but minimally.

damithc:
@m133225 if no objections, go ahead and merge this. In the meantime, we can keep a lookout for a better alternative. But delay merging if you want to research for a better alternative first.

m133225:
Actually, I think simply providing a unique id for GUI components should be sufficient. FxRobot can then find for `#id` node and cast it to the type required. Will look at other alternatives first.

damithc:
Yes, it should be sufficient for many cases, but not all. For example, how can we find things like the focused item in the list, the number of items in the list, if an item is faded, etc.?

damithc:
I see. Yes, that sounds workable. I'll give it a try.

damithc:
closing. No longer applicable.

Issue #357, opened 2016-07-01
Update documentations for Testing


m133225:

m133225:
Now that the testing is somewhat finalized, will work on this soon.

PR #391, opened 2016-07-03
Added instructions for running specific test classes in gradle

damithc:
as per title. This method worked for me. Not sure if there is an easier method though.

### DIFF
--- a/docs/Testing.md
+++ b/docs/Testing.md
@@ -15,6 +15,29 @@ There are a few key gradle tasks defined that we can play around with:
 - `./gradlew headless` to run **headless testing** only.  
  - Running this or `headful` repeatedly will not re-run the tests unless the build files are `clean`ed
 - `./gradlew headful` to run **headful testing** only.
+- Running specific test classes in a specific order:
+  - Create a test suite (to specify the test order). Here is an example:
+     ```java
+     package address;
+
+     import org.junit.runner.RunWith;
+     import org.junit.runners.Suite;
+
+     @RunWith(Suite.class)
+     @Suite.SuiteClasses({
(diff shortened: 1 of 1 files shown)


m133225:
I was considering this option, but I thought it would be better to first try and avoid leaks between tests. 

Also, it will be additional information for developers to keep track of - they have to add new tests to a test suite - which may be missed out over time.

docs/Testing.md
```diff
@@ -15,6 +15,29 @@ There are a few key gradle tasks defined that we can play around with:
...
+             address.guitests.PersonOverviewTest.class
+     })
+
+     public class TestsInOrder {
+         // the class remains empty,
+         // used only as a holder for the above annotations
+     }
+     ```
+  - Run the test suite using the gradle command <br>
+   `gradle clean headless --tests address.TestsInOrder`
```
m133225: This also works for individual test classes, not just for test suites. Planned to add these documentations in #357.

damithc:
This is for troubleshooting state leaks only. Not to be a part of permanent
testing.

-- 
Sent from a mobile device.

damithc:
I'll update the documentation to mention that this is for troubleshooting only.

docs/Testing.md
```diff
@@ -1,28 +1,54 @@
...
+
+
+## CI Testing
+- The current Travis set up
+  - runs the `./gradlew` command.
```
m133225: Currently we are running `./gradlew clean headless jacocoRootReport coveralls`.
Will update this documentation in a separate PR (as well as tweak the tasks so that it is easier to understand).

damithc:
OK, I'm merging this version then.

PR #393, opened 2016-07-04
Don't create additional jar files in the root folder #377

### LINKED ISSUE #377: Don't create additional jar files in the root folder
We should try to keep the root folder as clean as possible

HansNewbie:
Fixes #377

### DIFF
--- a/src/main/resources/updater/Instruction to use past versions.txt
+++ b/src/main/resources/updater/Instruction to use past versions.txt
@@ -1,1 +1,1 @@
-If the latest version does not work for some reason, you may want to use a past version until the latest version is fixed. In such a case, you can copy the JAR file of the version you want to use to the root folder (no need to rename it) and run that JAR file.
+If the latest version does not work for some reason, you may want to use a past version until the latest version is fixed. In such a case, you can copy the JAR file of the version you want to use from the past_versions folder to the root folder (no need to rename it) and run that JAR file.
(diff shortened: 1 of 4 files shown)


src/main/resources/updater/Instruction to use past versions.txt
```diff
@@ -0,0 +1 @@
+Copy the relevant version JAR to the root folder (no need to rename it) and run that JAR file.
```
HansNewbie: Is this instruction to use backup versions enough? What happens is all backup version will be in `past_versions/`. This `instruction.txt` will be in that folder as well.

To use a backup version, user just need to copy that JAR backup to the root and run that file.

damithc: `If the latest version does not work for some reason, you may want to use a past version until the latest version is fixed. In such a case, you can copy the JAR file of the version you want to use to the root folder (no need to rename it) and run that JAR file.`

HansNewbie: Should this instruction be put in the root folder or inside `past_versions/`? I am concerned that putting it in `past_versions/` might hide it from users, but putting it in the root makes the root folder dirty.

damithc: Yes, we can put it in the root folder too.

HansNewbie: Noted.

HansNewbie:
Will merge if no more comments.

PR #394, opened 2016-07-04
create separate gradle tasks for different type of testing #388

### LINKED ISSUE #388: create separate gradle tasks for different type of testing
May be we can use gradle TestFilter feature to create separate tasks for guitests, unittests, browsertest, etc.?

m133225:
Fixes #388

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -335,7 +335,7 @@ task guiUnitTest(type: AddressBookTest) {
 }
 
 task unitTest(type: AddressBookTest) {
-    include 'address/unittests/*'
+    include 'address/*'
     jacoco {
         destinationFile = new File("${buildDir}/jacoco/test.exec")
     }
--- a/build.gradle
+++ b/build.gradle
@@ -321,21 +321,21 @@ class AddressBookTest extends Test {
 }
 
 task guiTest(type: AddressBookTest) {
-    include 'guitests/*'
+    include 'guitests/**'
     jacoco {
         destinationFile = new File("${buildDir}/jacoco/test.exec")
     }
 }
 
 task guiUnitTest(type: AddressBookTest) {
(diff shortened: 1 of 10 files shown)


build.gradle
```diff
@@ -303,26 +306,58 @@ task coverage {
 task checkStyle {
 }
 
+ext {
+    guiTests = 'address/guitests/*'
+    guiUnitTests = 'address/guiunittests/*'
+    unitTests = 'address/unittests/*'
```
damithc: Is this fix easily extended to the scenario where we dismantle the unittests package to equalize test/functional packages?

m133225: Do you mean the case where we further sub-divide the unit tests? Then this should be extendable, unless the group we wish to test originates from various packages e.g. unittest/A and unittest/B under a fileRelatedTest task. In that case we have to use an array of filters instead.

damithc: No, I mean there will not be a unittests pacakge. instead of `address/unitests/events`, we'll have `address/events` (so that test classes are in the same package as the SUT).

m133225: Probably not. Need to tweak the settings so that it reads an array of filters instead.

m133225:
Will wait after refactoring of the test package structure to continue. Planning to separate each package into a separate travis build for faster testing.

damithc:
Understood. Do try to use wildcards though. If not, we have to modify build file every time we add a new package or rename one.

m133225:
@damithc Should we keep the 3 parallel builds? The speed up is probably not more than 1 minute unless there are more tests cases.

damithc:
I don't mind parallel builds, but I do want to keep things as close to the local builds as possible. The more different they are, harder to troubleshoot.

PR #396, opened 2016-07-04
Single file download for the app #378

### LINKED ISSUE #378: Single file download for the app
At the moment, new users have to download 2 jar files (am I correct?).
Ideally, it should be a single file.

HansNewbie:
Fixes #378 by putting all files to be downloaded in a release only for downloading purpose


damithc:
No comments. May be do another release after this is done. If this new version is not backward compatible, we need to do two releases to test the auto-updating.

PR #399, opened 2016-07-04
Change all FXML load errors dialog into assertion #288

### LINKED ISSUE #288: Change all FXML load errors dialog into assertion
As the FXMLs are suppose to be there and load without errors.

For example in this code segment
```
  public void showTagList(ObservableList<Tag> tags) {
        logger.debug("Loading tag list.");
        final String fxmlResourcePath = FXML_TAG_LIST;
        try {
            // Load the fxml file and create a new stage for the popup dialog.
            FXMLLoader loader = loadFxml(fxmlResourcePath);
            AnchorPane page = loader.load();

            Scene scene = new Scene(page);
            Stage dialogStage = loadDialogStage("List of Tags", primaryStage, scene);

            // Set the tag into the controller.
            TagListController tagListController = loader.getController();
            tagListController.setMainController(this);
            tagListController.setModelManager(modelManager);
            tagListController.setTags(tags);
            tagListController.setStage(dialogStage);

            // Show the dialog and wait until the user closes it
            dialogStage.showAndWait();
        } catch (IOException e) {
            logger.warn("Error loading tag list view: {}", e);
            showAlertDialogAndWait(AlertType.ERROR, "FXML Load Error", "Cannot load fxml for tag list.",
                                   "IOException when trying to load " + fxmlResourcePath);
        }
    }
```

yl-coder:
Fixes #288

### DIFF
--- a/src/main/java/address/controller/MainController.java
+++ b/src/main/java/address/controller/MainController.java
@@ -132,7 +132,7 @@ public class MainController extends UiController{
         } catch (IOException e) {
             logger.fatal("Error initializing root layout: {}", e);
             showFatalErrorDialogAndShutdown("FXML Load Error", "Cannot load fxml root layout.",
-                                         "IOException when trying to load ", fxmlResourcePath);
+                                            "IOException when trying to load ", fxmlResourcePath);
         }
     }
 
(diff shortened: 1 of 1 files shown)


YijinL:
Repeating almost the same code for every fxml view, maybe can abstract it out as a method (that takes in view name as a string for error message customisation).
Then call that method in all the `catch (IOException)` blocks. So all the `assert` and `showAlertDialogAndWait` are centralised.

src/main/java/address/controller/MainController.java
```diff
@@ -559,4 +545,10 @@ public void stop() {
         getPrimaryStage().hide();
         releaseResourcesForAppTermination();
     }
+
+    private void showAlertDialogWithAssertion(String title, String headerText, String contentText, String errorLocation) {
```
YijinL: Name is too general, this is specifically for fxml load failures right?
How about something like `handleFatalFxmlLoadError`?

YijinL:
no more comments

Issue #231, opened 2016-06-16
Enable assertions in the production release


damithc:
As per title

damithc:
@m133225 can you look into this?

YijinL:
If it's too hard to do, we can always write a utility assert function (`throws AssertionError`) to replace all `assert X : Y` statements

damithc:
That may be our last resort. We do that in TEAMMATES because app engine does not support assertions. If gradle can build the jar with assertions enabled, then we should use that.

m133225:
Actually, it seems like this can be done by simply adding `-ea` to the launch command in `Installer`.

Experimented with dependencies like [Capsule](https://github.com/danthegoodman/gradle-capsule-plugin) (still unable to figure out the correct way to use it with our installer x.x), but seems like it will not be needed.

damithc:
It's great if it can be done by adding `-ea` to the installer :+1:

HansNewbie:
@m133225 actually that `-ea` in `Installer` will not fix this issue. `Installer` will be discarded after the application has been unloaded, so the next time the application is opened, it won't be through `Installer` but directly to `MainApp`.

We really need to enable assertion through gradle build (I tried searching about it to no avail). Maybe we really need `throws AssertionError`.

Another solution is changed the way `Installer` works so that it will never get removed and the app will always be launched from there. But that also means the app will take about 5MB more of the disk space.

damithc:
I'm not sure it can be enabled through gradle. It's a runtime flag, not a compiletime flag. 5Mb extra space may be ok. 
Is it possible to have another class that launches the main app using a command (just like installer is doing now)?

HansNewbie:
Will think about how to approach this best.

PR #401, opened 2016-07-04
Not removing installer on updating

HansNewbie:
So that assertion is enabled in production. A proper fix to #231.


damithc:
No comments from me. Doesn't this require a change to dev docs? If not, then the dev docs are incomplete.

Issue #413, opened 2016-07-05
UI: Show the history of commands/actions


yl-coder:
To show history of commands/actions based on a observable list passed by modelManager.

PR #415, opened 2016-07-05
Model maintains command history #412

### LINKED ISSUE #412: Maintain history of completed/terminated commands
Should store as obslist so UI can show at status bar

YijinL:
Resolves #412 
Related: #413 

@yl-coder: calling `ModelManager.getFinishedCommands()` will return an unmodifiable observable list of `CommandInfo` elements which will be updated dynamically by model. Newly finished commands are appended to the end of the list

### DIFF
--- a/src/main/java/address/model/CommandInfo.java
+++ b/src/main/java/address/model/CommandInfo.java
@@ -1,0 +1,6 @@
+package address.model;
+
+public interface CommandInfo {
+    String getName();
+    String statusString();
+}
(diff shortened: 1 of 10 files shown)


src/main/java/address/model/CommandInfo.java
```diff
@@ -0,0 +1,6 @@
+package address.model;
+
+public interface CommandInfo {
+    String getName();
+    String statusString();
+}
```
YijinL: For UI status display

yl-coder:
Looks good. WIll be good if there is a test case to check the exposed methods. for e.g. check if the list is updated correctly after a few commands. I guess we can do that in the modelManagertest in future

YijinL:
Yeah will add in the tests in future

PR #417, opened 2016-07-05
Improve ui pending state command info #395 

### LINKED ISSUE #395: UI: Improve pending state message
State 3 different pending state. New, Deleted, Edited. With countdown in secs at the right side of the state. Show state and spinner after countdown reaches zero.

YijinL:
First half of #395 

@yl-coder: now `ReadOnlyViewablePerson` has a `getChangeInProgress()` method which returns the enum:
```java
enum ChangeInProgress { NONE, ADDING, EDITING, DELETING }
```
You can also listen to it's `ReadOnlyProperty` from the `changeInProgress()` method.

### DIFF
--- a/src/main/java/address/model/AddPersonCommand.java
+++ b/src/main/java/address/model/AddPersonCommand.java
@@ -1,6 +1,8 @@
 package address.model;
 
 import static address.model.ChangeObjectInModelCommand.State.*;
+import static address.model.datatypes.person.ReadOnlyViewablePerson.ChangeInProgress.*;
+
 import address.events.BaseEvent;
 import address.events.CreatePersonOnRemoteRequestEvent;
 import address.model.datatypes.person.Person;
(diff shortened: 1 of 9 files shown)


src/main/java/address/model/AddPersonCommand.java
```diff
@@ -82,7 +85,10 @@ protected void after() {
     protected State simulateResult() {
         assert input != null;
         // create VP and add to model
-        viewableToAdd = PlatformExecUtil.callAndWait(() -> model.addViewablePersonWithoutBacking(input), null);
+        PlatformExecUtil.runAndWait(() -> {
+            viewableToAdd = model.addViewablePersonWithoutBacking(input);
+            viewableToAdd.setChangeInProgress(ADDING);
```
yl-coder: Is addViewablePersonWithoutBacking use in other places? If not, i think  viewableToAdd.setChangeInProgress(ADDING); should be inside that method.

yl-coder:
Looks fine. no other comments.

src/main/java/address/model/AddPersonCommand.java
YijinL: No, but the status setting makes more sense existing in the command than the model code

damithc:
Branch can be deleted.

PR #419, opened 2016-07-05
Applied page object pattern to keybinding gui tests

damithc:
A prelim application of page object pattern

### DIFF
--- a/src/test/java/guitests/GuiTestBase.java
+++ b/src/test/java/guitests/GuiTestBase.java
@@ -6,6 +6,9 @@ import address.keybindings.KeyBinding;
 import address.keybindings.KeySequence;
 import address.model.datatypes.ReadOnlyAddressBook;
 import address.util.TestUtil;
+import guitests.guihandles.MainGuiHandle;
+import guitests.guihandles.MainMenuHandle;
+import guitests.guihandles.PersonListPanelHandle;
 import javafx.scene.input.KeyCode;
 import javafx.scene.input.KeyCodeCombination;
 import org.junit.After;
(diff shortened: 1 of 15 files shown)


damithc:
Guys, have a look before I continue. I guess we should move out the page objects (currently they are inner classes of GuiTestBase) into separate classes.

src/test/java/guitests/GuiTestBase.java
```diff
@@ -82,17 +87,109 @@ public FxRobot type(KeyCode... keyCodes) {
...
-            e.printStackTrace();
-        }
-    }
-
     protected void pushKeySequence(KeySequence keySequence) {
         push((KeyCodeCombination)keySequence.getKeyCombination());
         push((KeyCodeCombination)keySequence.getSecondKeyCombination());
     }
 
+    protected EditWindow use_PERSON_EDIT_ACCELERATOR() {
```
m133225: I think these should still remain as camel case.

m133225:
Looks good to me.

yl-coder:
Looks good to me also.

src/test/java/guitests/GuiTestBase.java
damithc: I decided to use the same string here because it is also used as the description field of the KeyBinding object. So this string is used in multiple places, and I didn't want to create multiple versions of it. I'll keep them as they are for now, but keep in mind as a potential refactoring.

damithc:
Intensified the application of PO pattern.

src/test/java/guitests/guihandles/EditPersonDialogHandle.java
```diff
@@ -0,0 +1,70 @@
...
+
+    public static final String TITLE = "Edit Person";
+    private String githubUserNameFieldId = "#githubUserNameField";
+    private String firstNameFieldId = "#firstNameField";
+    private String lastNameFieldId = "#lastNameField";
+    private String cityFieldId = "#cityField";
+    private String tagSearchFieldId = "#tagList";
+    private String cancelButtonText = "Cancel";
+
+    public EditPersonDialogHandle(GuiTestBase guiTestBase) {
```
damithc: I think instead of passing GuiTestBase, this should be given a reference to the FxRobot instance.

damithc:
Guess I'll merge this soon and continue refinements in a new PR.

PR #421, opened 2016-07-06
Determine a suitable initial size for the Stage #367

### LINKED ISSUE #367: Determine a suitable initial size for the Stage
As per title.

Continuation from #366, #333

yl-coder:
Fixes #367


src/main/java/address/ui/Ui.java
```diff
@@ -4,18 +4,23 @@
 import address.browser.BrowserManager;
 import address.controller.MainController;
 import address.model.ModelManager;
+import address.model.UserPrefs;
 import address.util.Config;
 import javafx.scene.control.Alert;
 import javafx.stage.Stage;
+import javafx.util.Pair;
+import org.apache.commons.lang3.tuple.ImmutablePair;
```
damithc: Any alternative for this? Are we using Apache.commons in other places?

yl-coder: No only this place. Perhaps I could write my own custom ScreenSize class and implement Serializable.

damithc: Yes, I'd rather minimize additional libraries unless we really need them.
BTW, don't you need to save the position as well?

src/main/java/address/util/ScreenSize.java
```diff
@@ -0,0 +1,29 @@
+package address.util;
+
+import java.io.Serializable;
+
+/**
+ * A Serializable class that contains the screen width and height.
+ */
+public class ScreenSize implements Serializable {
```
damithc: Is this screen size or window size?
In future, you might be adding more attributes to this class. e.g. x-position, y-position, width of panel, etc. May be want to give a more general name. e.g. GuiSettings

src/main/java/address/util/ScreenUtil.java
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * A utility class to contain methods related to the Screen.
+ */
+public final class ScreenUtil {
+
+    public static ScreenSize getRecommendedScreenSize() {
+        Rectangle2D primScreenBounds = Screen.getPrimary().getVisualBounds();
+        return new ScreenSize(primScreenBounds.getWidth() * 0.5, primScreenBounds.getHeight() * 0.7);
+    }
```
damithc: You can put this in the ScreenSize class?

src/main/java/address/model/UserPrefs.java
```diff
@@ -35,4 +40,12 @@ public String getSaveLocationString() {
     public boolean isSaveLocationSet() {
         return getSaveLocation() != null;
     }
+
+    public ScreenSize getScreenSize() {
+        return screenSize == null ? ScreenUtil.getRecommendedScreenSize() : screenSize;
```
damithc: `return screenSzie == null ? new ScreenSize() : screenSize;`
The default constructor will use default values.

yl-coder: noted. changed

src/main/java/address/util/ScreenUtil.java
yl-coder: Yes removed.

src/main/java/address/util/ScreenSize.java
yl-coder: Window size, implemented the x and y position.

PR #436, opened 2016-07-07
Improve functionality of CloudManipulator #427

### LINKED ISSUE #427: Improve functionality of CloudManipulator
It only has a few basic functionalities to add/edit data and/or delay/fail responses.
We should also add more specific functionalities such as targetting only sync-ups or sync-downs, as well as creating a more informative UI e.g. change delay response button's colour until the delaying is complete.

m133225:
Fixes #427

### DIFF
--- a/src/main/java/address/sync/RemoteService.java
+++ b/src/main/java/address/sync/RemoteService.java
@@ -388,7 +388,7 @@ public class RemoteService implements IRemoteService {
     private List<Person> convertToPersonList(List<CloudPerson> cloudPersonList) {
         List<Person> convertedList = new ArrayList<>();
         cloudPersonList.stream()
-                .forEach(CloudPerson -> convertedList.add(convertToPerson(CloudPerson)));
+                .forEach(cloudPerson -> convertedList.add(convertToPerson(cloudPerson)));
 
         return convertedList;
     }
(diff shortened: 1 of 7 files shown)


src/main/java/address/sync/RemoteService.java
```diff
@@ -388,7 +388,7 @@ private long getRateResetFromHeader(HashMap<String, String> header) {
     private List<Person> convertToPersonList(List<CloudPerson> cloudPersonList) {
         List<Person> convertedList = new ArrayList<>();
         cloudPersonList.stream()
-                .forEach(CloudPerson -> convertedList.add(convertToPerson(CloudPerson)));
+                .forEach(cloudPerson -> convertedList.add(convertToPerson(cloudPerson)));
 
         return convertedList;
```
YijinL:
```java
return cloudPersonList.stream()
                      .filter(this::convertToPerson)
                      .collect(Collectors.toCollection(Arraylist::new));
```
would be slightly more efficient; doing `ArrayList::add` n times means that there may be multiple resizings and recopying of the array. Also it feels nice and clearer to use more of the stream methods

m133225: Noted. `map` should be used here instead of `filter` tho 😛

YijinL: :P

damithc:
W.r.t naming, May be we can have something like this?:

```
[CloudSimulator] <---- [CloudSimulatorAdminApi] <---- [CloudSimulatorAdminGui]
```
The `CloudSimulatorAdminApi` can be used to control the cloud simulator programmatically, during automated testing.

PR #438, opened 2016-07-07
UI: Show the history of commands/actions #413

### LINKED ISSUE #413: UI: Show the history of commands/actions

yl-coder:
Fixes #413 

image


damithc:
Name it as `User Logs` -> `Activity History`? 
We can omit the `User Actions` part, to reduce non-functional decorations from the UI.

yl-coder:
Should the logs be more detail?

Some of the things that we could consider adding is
- The time of actions
- The intermediate steps: Edited Person X locally -> Edited Person X Remotely. etc
- TableView?

@damithc any suggestions or comments?

damithc:
yes, we should definitely add more details. Lets call it history, to avoid confusing with our own logging.
I don't think a table view is good. Different actions will need different details. So a list view is better where each card can have different format.

damithc:
May be we should use the same technique as the person list i.e. bind to model data.

yl-coder:
Implemented the same way as Person list. To allow greater customization in future.

yl-coder:
@YijinL Have a look at this. I made one line of change in the Model :P

yl-coder:
Going to merge this one to continue the next PR, any comments can be commented in the next PR.

PR #449, opened 2016-07-08
Hide ID in the Person card #434

### LINKED ISSUE #434: Hide ID in the Person card
From an address book perspective, it doesn't make sense to show ID. May be hide it in the UI and show it only when hovering over name?

yl-coder:
Fixes #434


damithc:
Does it show the id when hovering over the card/name?

yl-coder:
It will show when hovering over the name. Do you want it to appear be over the whole card?

damithc:
No, over the name is good enough.

PR #450, opened 2016-07-08
Support a 'view documentation' menu linked to F1 #309

### LINKED ISSUE #309: Support a 'view documentation' menu linked to F1
This feature will need to load a different version of the page based on if the product is stable, early access, or bleeding edge (i.e. master).

yl-coder:
Fixes #309 
image


damithc:
We should fine a good source of free icons.

src/main/java/address/controller/MainController.java
```diff
@@ -435,6 +437,29 @@ public void showTagList(ObservableList<Tag> tags) {
...
+            Scene scene = new Scene(page);
+            Stage dialogStage = loadDialogStage("Help", null, scene);
+            dialogStage.getIcons().add(getImage(ICON_HELP));
+            // Show the dialog and wait until the user closes it
+            dialogStage.showAndWait();
+        } catch (IOException e) {
+            logger.fatal("Error loading tag list view: {}", e);
+            showFatalErrorDialogAndShutdown("FXML Load Error", "Cannot load fxml for tag list.",
+                    "IOException when trying to load ", fxmlResourcePath);
+        }
```
damithc: Two things to improve:
1. try to avoid very tall `try` blocks (or any blocks). Most likely it can be narrowed to the first few lines?
2. This try-catch structure seems to be repetitive. Any way to abstract it out?

damithc:
Try to find a good help/userguide template. After you've found a few, we can decide which one to use.  It should be free, and hopefully using standard stuff like bootstrap and jquery.

damithc:
No more comments from me.

src/main/java/address/controller/MainController.java
```diff
@@ -116,29 +117,23 @@ private void showTipOfTheDay() {
...
-
-            // Show the scene containing the root layout.
-            Scene scene = new Scene(rootLayout);
-            scene.setOnKeyPressed(event -> raisePotentialEvent(new KeyBindingEvent(event)));
-            setMinSize();
-            setDefaultSize();
-            primaryStage.setScene(scene);
+        // Load root layout from fxml file.
+        FXMLLoader loader = loadFxml(fxmlResourcePath);
+        rootLayout = (VBox) loadLoader(loader, "Error initializing root layout");
```
damithc: Looks like FXMLLoader is a temp object. May be abstract both lines into a method?

yl-coder: Some part may need to use the loader object to get its controller.

PR #462, opened 2016-07-08
CloudSimulator should be able to handle concurrent requests #457

### LINKED ISSUE #457: CloudSimulator should be able to handle concurrent requests
If multiple persons are deleted at the same time, probably only one person will successfully be deleted on the cloud.

m133225:
Fixes #457


src/main/java/address/sync/cloud/CloudRateLimitStatus.java
```diff
@@ -97,7 +98,16 @@ void setQuotaResetTime(long quotaResetTime) {
...
+    public void useQuota(int amount) throws RejectedExecutionException {
+        if (quotaRemaining < amount) throw new RejectedExecutionException("No more API quota");
         quotaRemaining -= amount;
     }
+
+    public void useQuota() throws RejectedExecutionException {
+        useQuota(1);
+    }
+
+    public void replenishQuota() {
```
m133225: Shouldnt need this. Old mechanism to deduct quota should work, but this object needs to be volatile.

m133225:
Replaced by #472.

Issue #382, opened 2016-07-02
jkeymaster error in travis headless mode


damithc:
Doesn't seem to affect the test results though
```
address.guitests.PersonEditDialogTest > testUpdatePerson STANDARD_ERROR

    9534 [Thread-11] WARN com.tulskiy.keymaster.x11.X11Provider - Error: BadAccess (attempt to access private resource denied)

    9536 [Thread-11] WARN com.tulskiy.keymaster.x11.X11Provider - Error: BadAlloc (insufficient resources for operation)

    9536 [Thread-11] WARN com.tulskiy.keymaster.x11.X11Provider - Error: BadColor (invalid Colormap parameter)

    9536 [Thread-11] WARN com.tulskiy.keymaster.x11.X11Provider - Error: BadGC (invalid GC parameter)

    9537 [Thread-11] WARN com.tulskiy.keymaster.x11.X11Provider - Error: BadIDChoice (invalid resource ID chosen for this connection)

    9537 [Thread-11] WARN com.tulskiy.keymaster.x11.X11Provider - Error: BadName (named color or font does not exist)

    9537 [Thread-11] WARN com.tulskiy.keymaster.x11.X11Provider - Error: BadLength (poly request too large or internal Xlib length error)

    9537 [Thread-11] WARN com.tulskiy.keymaster.x11.X11Provider - Error: BadImplementation (server does not implement operation)

    9538 [Thread-11] WARN com.tulskiy.keymaster.x11.X11Provider - Error: 18

    9538 [Thread-11] WARN com.tulskiy.keymaster.x11.X11Provider - Error: 19

    9538 [Thread-11] WARN com.tulskiy.keymaster.x11.X11Provider - Error: 20

    9538 [Thread-11] WARN com.tulskiy.keymaster.x11.X11Provider - Error: 21

    9539 [Thread-11] WARN com.tulskiy.keymaster.x11.X11Provider - Error: 22

    9539 [Thread-11] WARN com.tulskiy.keymaster.x11.X11Provider - Error: 23

    9540 [Thread-11] WARN com.tulskiy.keymaster.x11.X11Provider - Error: 24

    9540 [Thread-11] WARN com.tulskiy.keymaster.x11.X11Provider - Error: 25

    9540 [Thread-11] WARN com.tulskiy.keymaster.x11.X11Provider - Error: 26

    9541 [Thread-11] WARN com.tulskiy.keymaster.x11.X11Provider - Error: 27

[...]

PR #468, opened 2016-07-09
jkeymaster error in travis headless mode #382

damithc:
as per title


damithc:
Looks like it's fixed now.

PR #474, opened 2016-07-11
Create default data file if missing #432

### LINKED ISSUE #432: Create default data file if missing
When the app is run for the first time, the data file will be missing. 
Instead of showing an error message, an empty file should be created.
While we are at it, choose a good name for the data file (`addressbook-data.xml`?) and put it in a separate folder (`/data`?)

m133225:
Fixes #432


damithc:
Is the default file name is still `test.xml`. If so, needs to change it too.

m133225:
Yes, that should be a simple change in `Config`'s default values.

PR #482, opened 2016-07-11
Repeated app resizing fails #456

### LINKED ISSUE #456: Repeated app resizing fails
At least on Mac, it will correctly maximize and reset to default size the first 2 times the key combination is pressed. 
However, from then on it will remain at the same default size (with some vertical movement every 2 times the keys are pressed).

Possible solution: http://stackoverflow.com/questions/18998671/set-height-and-width-of-stage-and-scene-in-javafx
```
protected void setDefaultSize() {
        primaryStage.setMaximized(false);
        primaryStage.setIconified(false);
        primaryStage.hide();
        primaryStage.show();
}
```
works, but causes flinching of the app.

m133225:
Fixes #456

Code does not seem to make sense, but works on Mac. Will further test on Windows machine at home.

### DIFF
--- a/src/main/java/address/controller/MainController.java
+++ b/src/main/java/address/controller/MainController.java
@@ -546,7 +546,7 @@ public class MainController extends UiController{
     @Subscribe
     private void handleResizeAppRequestEvent(ResizeAppRequestEvent event){
         logger.debug("Handling the resize app window request");
-        Platform.runLater(this::resizeWindow);
+        Platform.runLater(this::handleResizeRequest);
     }
 
     @Subscribe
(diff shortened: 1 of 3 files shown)


src/main/java/address/controller/MainController.java
```diff
@@ -575,9 +560,15 @@ private void minimizeWindow() {
         primaryStage.setMaximized(false);
     }
 
-    private void maximizeWindow() {
+    private void resizeWindow() {
+        logger.info("Resizing window");
+        if (OsDetector.isOnMac()) primaryStage.hide();
         primaryStage.setMaximized(true);
         primaryStage.setIconified(false);
+        if (OsDetector.isOnMac()) primaryStage.show();
```
damithc: May want to comment on why the code is like that.

src/main/java/address/controller/MainController.java
```diff
@@ -575,9 +560,29 @@ private void minimizeWindow() {
...
+                // on Mac, setMaximized seems to work like "setResize"
+                // isMaximized also does not seem to return the correct value
+                primaryStage.setMaximized(true);
+            } else {
+                primaryStage.setMaximized(!primaryStage.isMaximized());
+            }
+        }
+
+        logger.debug("Stage width: {}", primaryStage.getWidth());
+        logger.debug("Stage height: {}", primaryStage.getHeight());
```
damithc: I don't like the 'pregnant woman' shape (caused by short branch followed by deep branch) of the code, but not sure if it can be avoided.
```
\
/
\
 \
  \
   \
   /
  /
 /
/
```

PR #496, opened 2016-07-12
More detailed command completion info for UI #495

### LINKED ISSUE #495: More detailed command completion info for UI
Abstracted information needed for display:

1. unique command id
- string representation of type of command (eg. add person, delete person)
- enum type representing possible result states (successful, fail, cancelled)
- string representation of command target data object's type
- string representation of unique identifier of command target
- name of command target before the command was executed
- name of command target after the command finishes execution

UI can mix, match, and format the above pieces of data as needed. If the command target's name does not change due to command execution, fields 6 and 7 will be equal strings.

YijinL:
Resolves #495 

@yl-coder take a look at the `SingleTargetCommandResult` class

### DIFF
--- a/src/main/java/address/controller/ActivityHistoryCardController.java
+++ b/src/main/java/address/controller/ActivityHistoryCardController.java
@@ -8,9 +8,6 @@ import javafx.scene.layout.HBox;
 
 import java.io.IOException;
 
-/**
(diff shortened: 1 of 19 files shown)


yl-coder:
Looks Good

PR #497, opened 2016-07-12
Create a new release V1.3.0ea

m133225:
Making an official PR for easier reference.

### DIFF
--- a/src/main/java/address/MainApp.java
+++ b/src/main/java/address/MainApp.java
@@ -104,7 +104,7 @@ public class MainApp extends Application {
     }
 
     protected StorageManager initStorageManager(ModelManager modelManager, Config config, UserPrefs userPrefs) {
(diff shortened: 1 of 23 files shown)


m133225:
Never mind, seems like merging through command line is more appropriate as it can avoid merge commit messages.

damithc:
I think releases are supposed to happen on the `early-access` or `stable` branches. @HansNewbie confirm.

m133225:
Yes, this was a PR to merge `master` to `early-access`. Also, after merging on `early-access` and creating the release, I just simply merged back to master so that the version is updated on master

HansNewbie:
Yes. Releases should happen on the branches. Usually I don't merge back to
master, in which the version in master will be outdated. Dev is expected to
look at last commit message in updating the version (in my case).

Otherwise, either merge back to master or update the version before merging
branches (both are as complicated in terms of workflow).

PR #504, opened 2016-07-12
Refactor updater #503

### LINKED ISSUE #190: Disable updater in developer environment
Although it will not affect codes, it will affect creating release whereby unused libraries might be included due to updater testing.

### LINKED ISSUE #503: Refactor updater
and its components. 
Some segments are in big chunks which may be difficult for new developers to read and understand.

m133225:
Fixes #503.
Planning to also fix #190.

### DIFF
--- a/src/main/java/address/updater/BackupHandler.java
+++ b/src/main/java/address/updater/BackupHandler.java
@@ -111,15 +111,18 @@ public class BackupHandler {
         }
 
         List<String> backupFilesNames = getSortedBackupFilesNames();
-        List<String> backupFilesToDelete = getBackupFilesToDelete(backupFilesNames);
+        List<Version> backupVersions = getVersionsFromFileNames(backupFilesNames);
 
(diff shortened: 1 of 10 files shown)


src/main/java/address/updater/BackupHandler.java
```diff
@@ -152,14 +138,30 @@ public void cleanupBackups() {
             }
         });
 
-        dependencyHistoryHandler.cleanUpUnusedDependencies(unusedVersions);
+        dependencyHistoryHandler.cleanUpUnusedDependencies(deletedVersions);
+    }
+
+    private Version deleteBackupFile(String backupFileName) {
```
damithc: Try to avoid method that has side effects and returns a value at the same time. It's called the [Command-Query separation](https://en.wikipedia.org/wiki/Command%E2%80%93query_separation)

m133225:
Will also be looking at exception handling. Seems like there are many inappropriate use of exceptions and exception handling e.g. logging and rethrow

PR #505, opened 2016-07-13
Move test util files to testutils package #502

### LINKED ISSUE #502: Move test util files to testutils package
Some of the files in test/address/util should be moved to test/address/testuitls as the former is for test classes of util package and the latter is for utility classes used for testing.
At the same time, can rename testutils --> testutil to be consistent with util name (both singular).

yl-coder:
Fixes #502


yl-coder:
Will do this after PR#491 is merged

src/test/java/address/keybindings/KeyBindingsManagerApiTest.java
```diff
@@ -2,7 +2,7 @@
 
 
 import address.events.*;
-import address.util.TestUtil;
+import address.testutil.TestUtil;
```
damithc: We can put this in address.keybindings

src/test/java/address/TestApp.java
```diff
@@ -8,7 +8,7 @@
 import address.sync.cloud.model.CloudAddressBook;
 import address.util.Config;
 import address.util.GuiSettings;
-import address.util.TestUtil;
+import address.testutil.TestUtil;
```
damithc: Perhaps keep this in the same place MainApp is?

src/test/java/address/keybindings/KeyBindingsManagerApiTest.java
damithc: Never mind. My mistake.

PR #508, opened 2016-07-13
Finish PersonEditGuiTest and Gui Todos

yl-coder:
As per title


src/test/java/guitests/KeyBindingsGuiTest.java
```diff
@@ -15,13 +15,13 @@
...
-        ab.addPerson(new Person("Person3", "Lastname3", 3));
-        ab.addPerson(new Person("Person4", "Lastname4", 4));
-        ab.addPerson(new Person("Person5", "Lastname5", 5));
+        ab.addPerson(new Person("John", "Wilson", 1));
+        ab.addPerson(new Person("John", "Lennon", 2));
+        ab.addPerson(new Person("Allan", "Turing", 3));
+        ab.addPerson(new Person("Obama", "Micheal", 4));
+        ab.addPerson(new Person("Pedo", "Lee", 5));
         return ab;
     }
```
damithc: Use the same set of `typical data` (used in PersonEditGuitTest) here as well? Also note the naming convention used there. It's not totally random.

src/test/java/guitests/guihandles/PersonListPanelHandle.java
```diff
@@ -105,7 +109,7 @@ public void use_PERSON_DELETE_ACCELERATOR() {
     }
 
     public FxRobot waitForGracePeriodToExpire() {
-        return guiRobot.sleep(4000);//TODO: tie the sleep duration to the actual grace period and implement a polling wait
+        return guiRobot.sleep((ModelManager.GRACE_PERIOD_DURATION + 1) * 1000);//TODO: Implement a polling wait
```
damithc: Create a `sleepForGracePeriod()` method in GuiTestBase?

src/test/java/guitests/KeyBindingsGuiTest.java
yl-coder: Noted. using the TypicalData.book now

src/test/java/guitests/guihandles/PersonListPanelHandle.java
yl-coder: ok

yl-coder:
@damithc For testing using edit button and edit context menu, Do we just test if the edit dialog open successfully? Since we already tested the rest. Or to test the whole scenario again?

yl-coder:
@damithc What do you mean by these TODOS at PersonEditGuiTest.
    /* TODO:
     * Test 'OK'
     * Test data validation (just one case)
     */

damithc:
Test OK: click ok button rather than press enter
data validation: if the entered values are not acceptable e.g. last name missing, wrong date format etc.

yl-coder:
After looking through the screenshots, I can conclude that most of the assertion methods seems to be working fine, just that some actions method is not working properly:
Here's the summary problematic ones in headfull
guiRobot.push, Accelerator like Delete and jump to index is not working properly. 
clickOn, erase, and write.(Maybe all of them or one of them, since they are normally use in sequences)

And also there are cases where the window is not appearing on the screen, which cause some assertion to fail.

damithc:
Try to find workarounds for the ones that don't work.
Delete is an accelerator while jump to index is a shortcut, so they are not exactly similar. If both are not working, the problem seem to be related to simulating keys.

damithc:
Let's keep at it. It's going to take a lot of trial-and-error before we can find some workable solutions.

yl-coder:
Gonna setup travis locally, hopefully can find the root cause.

src/test/java/guitests/GuiRobot.java
```diff
@@ -42,4 +44,35 @@ public void pushKeySequence(KeySequence keySequence) {
...
+        int count = 0;
+        while (count < 10) {
+            try {
+                return super.clickOn(query, buttons);
+            } catch (FxRobotException e) {
+                sleep(500);
+                count++;
+            }
+        }
+        return super.clickOn(query, buttons);
```
damithc: Either don't return anything or return GuiRobot? Otherwise the chained method call will go to FxRobot even if GuiRobot overrides it.

yl-coder: If i remembered, if an instance originally is a GuiRobot, even if cast to other sub or parent class, it will still go to the GuiRobot method

damithc: That's true. But the caller will not be able to chain any new methods added in GuiRobot.

yl-coder: true, will change all methods in guiRobot to return GuiRobot

.travis.yml
```diff
@@ -10,5 +10,4 @@ before_install:
 addons:
   apt:
     packages:
-      - oracle-java8-installer
-
+      - oracle-java8-installer
```
damithc: Is this change needed?

yl-coder: No, I guess it is created by merge conflict

src/main/java/address/controller/MainController.java
```diff
@@ -44,6 +44,9 @@
  * The controller that creates the other controllers
  */
 public class MainController extends UiController{
+    public static final String STAGE_TITLE_TAG_SELECTION = "Tag Selection";
+    public static final String STAGE_TITLE_MANAGE_TAG_DIALOG = "List of Tags";
+    public static final String STAGE_TITLE_NEW_TAG = "New Tag";
```
damithc: Let's name dialogs consistently. Perhaps by the action use is performing? e.g. `Tagging a Person`, `Adding a New Tag`, `Listing All Tags` etc.

src/main/java/address/controller/MainController.java
```diff
@@ -608,8 +614,8 @@ private void resizeWindow() {
             primaryStage.setMaximized(!primaryStage.isMaximized());
         }
 
-        logger.debug("Stage width: {}", primaryStage.getWidth());
-        logger.debug("Stage height: {}", primaryStage.getHeight());
+        logger.debug("After: Stage width: {}", primaryStage.getWidth());
+        logger.debug("After: Stage height: {}", primaryStage.getHeight());
```
damithc: May be this two lines can be combined?

src/main/java/address/controller/TagListController.java
```diff
@@ -21,9 +21,9 @@
     ModelManager modelManager;
 
     @FXML
-    private AnchorPane mainPane;
+    private AnchorPane tagListMainPane;
     @FXML
-    private ScrollPane tags;
+    private ScrollPane tagListTags;
```
damithc: It is hard to distinguish between these two objects based on their names. Both are Panes but you used `Pane` in only one name.

src/main/java/address/controller/MainController.java
yl-coder: I guess by the a simplified controller name that controls the dialog will be better.

src/test/java/guitests/GuiRobot.java
```diff
@@ -5,41 +5,74 @@
...
+        //Busy waiting implementation to fix issue when app window is not brought up to screen yet.
+        int count = 0;
+        while (count < 10) {
+            try {
+                return (GuiRobot) super.clickOn(query, buttons);
+            } catch (FxRobotException e) {
+                sleep(500);
+                count++;
+            }
+        }
```
damithc: Can be abstract the busy waiting to another method using Lambdas? Then it can be pushed to parent class.

src/test/java/guitests/GuiTestBase.java
```diff
@@ -83,11 +104,20 @@ protected String getDataFileLocation() {
...
+    }
+
     @After
     public void cleanup() throws TimeoutException {
+        File file = GuiTest.captureScreenshot();
+        try {
+            Files.copy(file, new File(this.getClass().getName() + name.getMethodName() + ".png"));
+        } catch (IOException e1) {
+            e1.printStackTrace();
+        }
```
damithc: File copying can be moved to a util class.

src/test/java/guitests/KeyBindingsGuiTest.java
```diff
@@ -1,103 +1,106 @@
...
 
     @Test
     public void keyBindings() {
-
         //======= shortcuts =======================
 
         personListPanel.use_LIST_ENTER_SHORTCUT();
-        assertTrue(personListPanel.isSelected("Person1", "Lastname1"));
+        assertTrue(personListPanel.isSelected("Alice", "Brown"));
```
damithc: Write a `isSelected(Person)` instead.

damithc: Even better if you can do something like this:
```java
aliceCard = getCard(alice);
//use the keybinding
assertEquals(aliceCard, getSelectedCard());
```

src/main/java/address/controller/MainController.java
yl-coder: ok sure

src/test/java/guitests/guihandles/GuiHandle.java
```diff
@@ -1,12 +1,14 @@
 package guitests.guihandles;
 
-
+import address.util.LoggerManager;
+import com.google.common.base.Optional;
```
damithc: Why use Google's class instead of Java built-in?

src/test/java/guitests/guihandles/GuiHandle.java
```diff
@@ -28,15 +33,39 @@ public GuiHandle(GuiRobot guiRobot, Stage primaryStage) {
...
+
+        if(!window.isPresent()) {
+            LoggerManager.getLogger(this.getClass()).fatal("Can't find stage " + stageTitle + ", Therefore, aborting focusing");
+            return;
+        }
+
+        guiRobot.targetWindow(window.get());
+        guiRobot.interact(() -> window.get().requestFocus());
+        LoggerManager.getLogger(this.getClass()).info("Finishing focus " + stageTitle);
+    }
```
damithc: Looks like this should be a method of `mainGui` (as it involves all windows shown at that moment)?

src/main/java/address/controller/TagListController.java
yl-coder: Right, previously was a hotfix to solve to unique fx id, as tags is too generic.

src/test/java/guitests/guihandles/MainGuiHandle.java
```diff
@@ -49,4 +59,25 @@ public void use_APP_RESIZE_HOTKEY() {
         guiRobot.sleep(1000);
     }
 
+    public FxRobot sleepForGracePeriod() {
+        return guiRobot.sleep((ModelManager.GRACE_PERIOD_DURATION + 1), TimeUnit.SECONDS);
+    }
+
+    public FxRobot sleep(long duration, TimeUnit timeunit) {
+        return guiRobot.sleep(duration, timeunit);
+    }
```
damithc: Why do you need this here? If at all, it should be in GuiHandle class, right?

src/test/java/guitests/guihandles/GuiHandle.java
yl-coder: TestFx uses them :P.

src/test/java/guitests/guihandles/MainGuiHandle.java
```diff
@@ -49,4 +59,25 @@ public void use_APP_RESIZE_HOTKEY() {
...
+        if(!window.isPresent()) {
+            LoggerManager.getLogger(this.getClass()).fatal("Can't find Main App Stage");
+            throw new RuntimeException("Can't find Main App Stage");
+        }
+
+        guiRobot.targetWindow(window.get());
+        guiRobot.interact(() -> window.get().requestFocus());
+        LoggerManager.getLogger(this.getClass()).debug("Focused target window name: "
+                + ((Stage)guiRobot.targetWindow()).getTitle());
+    }
```
damithc: Some duplication with another method above?

src/test/java/guitests/guihandles/ManageTagsDialogHandle.java
```diff
@@ -1,14 +1,66 @@
...
     public ManageTagsDialogHandle(GuiRobot guiRobot, Stage primaryStage) {
-        super(guiRobot, primaryStage);
+        super(guiRobot, primaryStage, MainController.STAGE_TITLE_MANAGE_TAG_DIALOG);
+    }
+
+    private ScrollPane getScrollPane() {
+        return (ScrollPane) getNode("#tagListTags");
+    }
+
+    public List<String> getListOfTagNames() {
```
damithc: getTagNames? The `List` part is implied by return type.

src/test/java/guitests/guihandles/ManageTagsDialogHandle.java
```diff
@@ -1,14 +1,66 @@
...
+
+    public static final String EDIT_TAG_TEXT_FIELD = "#tagNameField";
+
     public ManageTagsDialogHandle(GuiRobot guiRobot, Stage primaryStage) {
-        super(guiRobot, primaryStage);
+        super(guiRobot, primaryStage, MainController.STAGE_TITLE_MANAGE_TAG_DIALOG);
+    }
+
+    private ScrollPane getScrollPane() {
+        return (ScrollPane) getNode("#tagListTags");
```
damithc: To be consistent, use a constant instead of magic string.

src/test/java/guitests/guihandles/ManageTagsDialogHandle.java
```diff
@@ -1,14 +1,66 @@
...
+
+    }
+
+    public boolean contains(String value) {
+        return getListOfTagNames().stream().filter(value::equals).findAny().isPresent();
     }
 
+    public void openEditTagDialog(String tag) {
+        guiRobot.doubleClickOn(tag);
+        focusOnWindow("Edit Tag");
```
damithc: Magic string.

src/test/java/guitests/guihandles/ManageTagsDialogHandle.java
```diff
@@ -1,14 +1,66 @@
...
+        super(guiRobot, primaryStage, MainController.STAGE_TITLE_MANAGE_TAG_DIALOG);
+    }
+
+    private ScrollPane getScrollPane() {
+        return (ScrollPane) getNode("#tagListTags");
+    }
+
+    public List<String> getListOfTagNames() {
+        ObservableList<Node> childrenUnmodifiable = ((VBox) getScrollPane().getContent()).getChildren();
+        return childrenUnmodifiable.stream().map(n -> ((Label) n.lookup("#tagName")).getText())
```
damithc: Magic string.

src/test/java/guitests/guihandles/ManageTagsDialogHandle.java
```diff
@@ -1,14 +1,66 @@
...
+        return getTextFieldText(EDIT_TAG_TEXT_FIELD);
+    }
+
+    public void changeEditTagDialogText(String text) {
+        guiRobot.clickOn(EDIT_TAG_TEXT_FIELD).push(KeyCode.SHORTCUT, KeyCode.A).eraseText(1).write(text);
+    }
+
+    public boolean isChangeEditTagDialogOpen() {
+        try{
+            return guiRobot.window("Edit Tag") != null;
```
damithc: Magic string.

src/test/java/guitests/GuiRobot.java
yl-coder: The parent class is FxRobot. Can't shift it there. Perhaps i can extract it to a internal private method.

src/test/java/guitests/guihandles/PersonCardHandle.java
```diff
@@ -49,8 +75,27 @@ public boolean isSamePerson(Person person){
     }
 
     @Override
+    public boolean equals(Object obj) {
+        if (obj instanceof Person) {
+            Person person = (Person) obj;
+            return getFirstName().equals(person.getFirstName())
+                    && getLastName().equals(person.getLastName())
+                    && getAddress().equals(PersonCardController.getAddressString(person.getStreet(),
+                    person.getCity(), person.getPostalCode()));
```
damithc: Improve indentation for better readability.

src/test/java/guitests/guihandles/PersonCardHandle.java
```diff
@@ -49,8 +75,27 @@ public boolean isSamePerson(Person person){
...
+        }
+
+        if(obj instanceof PersonCardHandle) {
+            PersonCardHandle handle = (PersonCardHandle) obj;
+            return getFirstName().equals(handle.getFirstName())
+                    && getLastName().equals(handle.getLastName())
+                    && getAddress().equals(handle.getAddress()) && getBirthday().equals(handle.getBirthday());
```
damithc: This line break is at an odd place.

src/test/java/guitests/guihandles/PersonListPanelHandle.java
```diff
@@ -54,38 +78,60 @@ public ReadOnlyViewablePerson getSelectedPerson() {
         return (ListView<ReadOnlyViewablePerson>) getNode(PERSON_LIST_VIEW_ID);
     }
 
+    /**
+     * Clicks on the middle of the Listview.
+     * In order for headfull testing to work in travis ci, listview needs to be clicked before firing hot keys.
+     */
+    public void clickOnListView() {
+        double x = getListView().localToScreen(getListView().getLayoutBounds()).getMinX() + getListView().getWidth() / 2;
+        double y = getListView().localToScreen(getListView().getLayoutBounds()).getMinY() + getListView().getHeight() / 2;
```
damithc: Calculating the midpoint can be abstracted and reused.

src/test/java/guitests/guihandles/PersonListPanelHandle.java
```diff
@@ -54,38 +78,60 @@ public ReadOnlyViewablePerson getSelectedPerson() {
...
-                guiRobot.push(KeyCode.SHORTCUT, KeyCode.DIGIT7);
+                guiRobot.push(TestUtil.scrub(new KeyCode[]{KeyCode.SHORTCUT, KeyCode.DIGIT7}));
                 break;
             case 8:
-                guiRobot.push(KeyCode.SHORTCUT, KeyCode.DIGIT8);
+                guiRobot.push(TestUtil.scrub(new KeyCode[]{KeyCode.SHORTCUT, KeyCode.DIGIT8}));
                 break;
             case 9:
-                guiRobot.push(KeyCode.SHORTCUT, KeyCode.DIGIT9);
+                guiRobot.push(TestUtil.scrub(new KeyCode[]{KeyCode.SHORTCUT, KeyCode.DIGIT9}));
```
damithc: Revert experimental changes that did not help.

src/test/java/guitests/guihandles/PersonListPanelHandle.java
```diff
@@ -119,20 +161,55 @@ public void navigateDown() {
...
         guiRobot.clickOn(person.getFirstName());
     }
 
+    /**
+     * Right click on Person to show context menu.
+     * @param person
+     */
+    public void rightClickOnPerson(Person person) {
+        //Instead of using guiRobot.rightCickOn(), We will be firing off contextmenu request manually.
+        //As there is a bug in monocle that doesn't show contextmenu by actual right clicking.
```
damithc: Give link to bug report.

src/test/java/guitests/GuiRobot.java
yl-coder: I don't think we can use lamdas for this

src/test/java/guitests/GuiTestBase.java
yl-coder: ok moved

src/test/java/guitests/guihandles/PersonListPanelHandle.java
```diff
@@ -145,17 +222,114 @@ public String getFilterText() {
...
         guiRobot.drag(firstNameOfPersonToDrag).dropTo(firstNameOfPersonToDropOn);
     }
 
     /**
+     * Drags the person cards outside of the listview.
+     * @param listOfPersonsToDrag The texts which identify the cards to be dragged.
+     */
+    public void dragOutsideList(List<String> listOfPersonsToDrag) {
+        double posY = this.getListView().localToScene(this.getListView().getBoundsInLocal()).getMaxY()
+                - 50;
```
damithc: Improve line wrapping.
```java
double posY = this.getListView().localToScene(this.getListView().getBoundsInLocal()).getMaxY() 
              - 50;
```
Use explanatory variable for `this.getListView().localToScene(this.getListView().getBoundsInLocal()).getMaxY()`?

src/test/java/guitests/guihandles/PersonListPanelHandle.java
```diff
@@ -145,17 +222,114 @@ public String getFilterText() {
...
+        double y = this.primaryStage.getScene().getY() + this.primaryStage.getScene().getHeight() + 10;
+        guiRobot.drag(personToDrag).dropTo(x, y);
+    }
+
+    /**
+     * Drags cards outside of the Test App.
+     * @param listOfPersonsToDrag The texts which identify the cards to be dragged.
+     */
+    public void dragOutsideApp(List<String> listOfPersonsToDrag) {
+        double x = this.primaryStage.getScene().getX() + this.primaryStage.getScene().getWidth() + 10;
```
damithc: Abstract these calculations into small methods and reuse them.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -42,35 +46,143 @@ public void editPerson() {
...
 
-        assertTrue(descOnFail(alicePersonCard, newAlice), alicePersonCard.isSamePerson(newAlice));
-        //TODO: write a more elegant function that uses assertEquals but compare two different types of objects
+        mainGui.sleep(1, TimeUnit.SECONDS);
+
+        //Confirm pending state correctness
+        assertTrue(alicePersonCard.isPendingStateRootVisible());
+        assertTrue(alicePersonCard.isPendingStateLabelVisible());
+        assertFalse(alicePersonCard.isPendingStateProgressIndicatorVisible());
+        assertTrue(alicePersonCard.getPendingStateLabel().equals("Editing"));
```
damithc: Lump these into a method in the GuiHandle. e.g. 
`assertTrue(alicePersonCard.isShowingGracePeriod("Editing")`

It's best when we can write test at higher level abstractions.

src/test/java/guitests/KeyBindingsGuiTest.java
yl-coder: ok

src/test/java/guitests/PersonEditGuiTest.java
damithc: Ideally, you should confirm that no other card is showing grace period.

src/test/java/guitests/guihandles/GuiHandle.java
```diff
@@ -28,15 +33,39 @@ public GuiHandle(GuiRobot guiRobot, Stage primaryStage) {
...
-            Object object = ctor.newInstance(new Object[] { guiRobot, primaryStage });
+            Object object = ctor.newInstance(new Object[] { guiRobot, primaryStage});
             return (T)object;
         } catch (Exception e) {
             throw new RuntimeException("Cannot create gui handle of type " + clazz.getName(), e);
         }
     }
 
+    public void focusOnWindow(String stageTitle) {
+        LoggerManager.getLogger(this.getClass()).info("Focusing " + stageTitle);
```
m133225: Logger could extracted into a constant. It should still keep the reference to the calling object's type. (See `ChangeObjectInModelCommand`)

src/test/java/guitests/guihandles/MainGuiHandle.java
yl-coder: I don't want all handle to be able to call sleep. Otherwise, we will handle personList.sleep(), statusBar.sleep(), etc. all over the testing codes.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -42,35 +46,143 @@ public void editPerson() {
...
+
+        //Confirm the right card is selected after the edit
+        assertEquals(alicePersonCard, newAlice);
+        assertEquals(1, personListPanel.getSelectedCards().size());
+        assertEquals(alicePersonCard, personListPanel.getSelectedCards().get(0));
```
damithc: These lines too can be abstracted into a method in *GuiHandle class.
e.g. `assertTrue(personListPanel.isSelectedOnly(alice))`

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -42,35 +46,143 @@ public void editPerson() {
...
+        assertEquals(newAlice.toString(), personListPanel.getSelectedPerson().toString());
+
+        //Confirm other cards are unaffected.
+        personListPanel.clickOnListView();
+        personListPanel.use_LIST_JUMP_TO_INDEX_SHORTCUT(2); //To make sure the card is visible
+        assertEquals(personListPanel.getPersonCardHandle(1), td.benson);
+        personListPanel.use_LIST_JUMP_TO_INDEX_SHORTCUT(3);
+        assertEquals(personListPanel.getPersonCardHandle(2), td.charlie);
+        personListPanel.use_LIST_JUMP_TO_INDEX_SHORTCUT(4);
+        assertEquals(personListPanel.getPersonCardHandle(3), td.dan);
```
damithc: Abstract into one method in *GuiHandle class.

src/test/java/guitests/guihandles/MainGuiHandle.java
yl-coder: Right, change to focusOnWindow(TestApp.APP_TITLE)

src/test/java/guitests/guihandles/ManageTagsDialogHandle.java
yl-coder: ok

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -42,35 +46,143 @@ public void editPerson() {
...
+        EditPersonDialogHandle editPersonDialog =  personListPanel.clickEdit();
+        editPersonDialog.enterFirstName("Peter");
+        editPersonDialog.enterLastName("");
+        editPersonDialog.clickOk();
+
+        assertTrue(editPersonDialog.isInputValidationErrorDialogShown());
+        editPersonDialog.dissmissErrorMessage("Invalid Fields");
+        Assert.assertFalse(editPersonDialog.isInputValidationErrorDialogShown());
+        
+    }
```
damithc: This method has extra blank lines.

src/test/java/guitests/guihandles/ManageTagsDialogHandle.java
yl-coder: ok

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -42,35 +46,143 @@ public void editPerson() {
...
+
+        personListPanel.clickOnPerson(td.alice);
+        EditPersonDialogHandle editPersonDialog =  personListPanel.clickEdit();
+        editPersonDialog.enterFirstName("Peter");
+        editPersonDialog.enterLastName("");
+        editPersonDialog.clickOk();
+
+        assertTrue(editPersonDialog.isInputValidationErrorDialogShown());
+        editPersonDialog.dissmissErrorMessage("Invalid Fields");
+        Assert.assertFalse(editPersonDialog.isInputValidationErrorDialogShown());
```
damithc: `Assert.`?

src/test/java/guitests/guihandles/ManageTagsDialogHandle.java
yl-coder: ok

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -42,35 +46,143 @@ public void editPerson() {
...
+        
+    }
+
+    @Test
+    public void cancelPerson_usingAccelerator() {
```
damithc: `cancelPerson`?

src/test/java/guitests/guihandles/ManageTagsDialogHandle.java
yl-coder: ok

damithc:
Added some comments. Didn't review the whole code though. I'll review again after revisions.

src/test/java/guitests/guihandles/ManageTagsDialogHandle.java
yl-coder: ok

src/test/java/guitests/guihandles/PersonCardHandle.java
yl-coder: ok

src/test/java/guitests/guihandles/PersonCardHandle.java
yl-coder: ok

src/test/java/guitests/guihandles/PersonListPanelHandle.java
yl-coder: ok abstracted

src/test/java/guitests/guihandles/PersonListPanelHandle.java
yl-coder: I guess we still need to scrub for platform compatibility?

src/test/java/guitests/guihandles/PersonListPanelHandle.java
yl-coder: ok

src/test/java/guitests/guihandles/PersonListPanelHandle.java
yl-coder: ok

src/test/java/guitests/guihandles/PersonListPanelHandle.java
yl-coder: ok

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -42,35 +46,136 @@ public void editPerson() {
...
+        mainGui.sleep(ModelManager.GRACE_PERIOD_DURATION/2, TimeUnit.SECONDS);
+        personListPanel.use_PERSON_CHANGE_CANCEL_ACCELERATOR();
+        assertEquals(alicePersonCard, td.alice);
+        assertEquals(statusBar.getText(), "Edit Person [ Alice Brown -> Alicia Brownstone ] was cancelled.");
+
+        //New
+        EditPersonDialogHandle addPersonDialog = personListPanel.clickNew();
+        Person pandaWong = new PersonBuilder(td.alice.copy()).withFirstName("Panda").withLastName("Wong")
+                .withStreet("Chengdu Panda Street").withCity("Chengdu").withPostalCode("PANDA")
+                .withBirthday("01.01.1979").withGithubUsername("panda").withTags(td.colleagues, td.friends).build();
```
damithc: This builder method is used for creating a person similar to an existing person. We can write a separate builder method for creating an entirely new person.

src/test/java/guitests/guihandles/MainGuiHandle.java
damithc: Why not put it in the GuiTestBase class? I think it is already there.

src/test/java/guitests/guihandles/PersonListPanelHandle.java
damithc: Yes. keep if necessary.

src/test/java/guitests/GuiRobot.java
damithc: I feel there must be a way to avoid repeating this loop, but yes, probably not a straightforward use of Lambdas.

src/test/java/guitests/PersonEditGuiTest.java
yl-coder: In that case, We couldn't make use of the assertEquals method? As this method give more details on the assert failure.

src/test/java/guitests/guihandles/MainGuiHandle.java
yl-coder: There is a TODO that states "//TODO: remove this from here, only *Handle objects should use the robot"
So there is no guiRobot in GuiTestBase now.

src/test/java/address/testutil/TestUtil.java
```diff
@@ -10,19 +10,26 @@
 import address.sync.cloud.model.CloudAddressBook;
 import address.sync.cloud.model.CloudPerson;
 import address.sync.cloud.model.CloudTag;
+import com.google.common.io.Files;
```
damithc: Is this the best Files util out there? If it is, it's good to know.

damithc: In which library did this file come in?

src/test/java/guitests/KeyBindingsGuiTest.java
```diff
@@ -1,103 +1,106 @@
...
         editPersonDialog.clickCancel();
 
+        mainGui.focusOnMainApp();
+        personListPanel.clickOnPerson(td.dan);
+
         personListPanel.use_PERSON_DELETE_ACCELERATOR();
-        assertTrue(personListPanel.contains("Person4", "Lastname4")); // still in the list due to grace period
-        personListPanel.waitForGracePeriodToExpire();
-        assertFalse(personListPanel.contains("Person4", "Lastname4")); // removed from list after grace period
+        assertTrue(personListPanel.contains("Dan", "Edwards")); // still in the list due to grace period
```
damithc: Pass the whole object rather than just the two names?

src/test/java/address/testutil/TestUtil.java
yl-coder: com.google.guava

yea, better than java File Util. Does all the dirty jobs for u.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -42,35 +45,133 @@ public void editPerson() {
...
+        assertEquals(newAlice.toString(), personListPanel.getSelectedPerson().toString());
+
+        //Confirm other cards are unaffected.
+        personListPanel.clickOnListView();
+        personListPanel.use_LIST_JUMP_TO_INDEX_SHORTCUT(2); //To make sure the card is visible
+        assertEquals(personListPanel.getPersonCardHandle(1), td.benson); //Ensure the card details are correct
+        personListPanel.use_LIST_JUMP_TO_INDEX_SHORTCUT(3);
+        assertEquals(personListPanel.getPersonCardHandle(2), td.charlie);
+        personListPanel.use_LIST_JUMP_TO_INDEX_SHORTCUT(4);
+        assertEquals(personListPanel.getPersonCardHandle(3), td.dan);
```
damithc: Instead, use a single method:
`assertTrue(personListPanel.isListMatching(person_list_expected))`

damithc: When it is just one line like this, you can call it in many other places to ensure unrelated cards are not affected by an operation.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -42,35 +45,133 @@ public void editPerson() {
...
+        assertTrue(deletedCard.isShowingGracePeriod("Deleting"));
+
+        personListPanel.use_PERSON_CHANGE_CANCEL_ACCELERATOR();
+        mainGui.sleep(1, TimeUnit.SECONDS);
+        assertFalse(deletedCard.isShowingGracePeriod("Deleting"));
+        assertEquals(statusBar.getText(), "Delete Person [ " + deletedCard.getFirstName() + " "
+                     + deletedCard.getLastName() + " ] was cancelled.");
+
+
+        //Edit
```
damithc: Too many blank lines. One is enough.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -42,35 +45,133 @@ public void editPerson() {
...
+        editPersonDialog.dissmissErrorMessage("Invalid Fields");
+        assertFalse(editPersonDialog.isInputValidationErrorDialogShown());
+    }
+
+    @Test
+    public void cancelOperation_usingAccelerator() {
+
+        //Delete
+        personListPanel.clickOnPerson(td.alice);
+        PersonCardHandle deletedCard = personListPanel.getPersonCardHandle(new Person(personListPanel.getSelectedPerson()));
```
damithc: The card is not deleted yet. Call it `aliceCard`?

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -42,35 +45,133 @@ public void editPerson() {
...
+        personListPanel.use_PERSON_DELETE_ACCELERATOR();
+        mainGui.sleep(1, TimeUnit.SECONDS);
+        assertTrue(deletedCard.isShowingGracePeriod("Deleting"));
```
damithc: The sleep can be inside the `isShowingGracePeriod`? Otherwise you have to repeat the sleep every time. When you find yourself repeating a code block, it's as sign that you are missing a higher level abstraction.

src/test/java/guitests/PersonEditGuiTest.java
damithc: Can you collapse the two lines like this?
`PersonCardHandle aliceCard = personListPanel.selectCard(td.alice);`

src/test/java/guitests/PersonEditGuiTest.java
yl-coder: Not all scenario we will need the sleep for 1 secs. The intention and purpose of sleeping for 1 secs is different from isShowingGracePeriod

src/test/java/guitests/guihandles/MainGuiHandle.java
damithc: See if you can put the sleep method in the GuiTestBase class and call the MainGuiHandle from there. That way, your code will read `sleep()` instead of `mainGui.sleep()`

src/test/java/guitests/guihandles/PersonListPanelHandle.java
```diff
@@ -145,17 +227,110 @@ public String getFilterText() {
...
+    }
+
+    /**
+     * Drags card to the top or bottom edge of the listview.
+     * @param dragFrom The text which identify the card to be dragged.
+     * @param direction To the top or bottom edge of the listview.
+     * @param dragDuration Drag duration
+     * @param timeunit Timeunit for the duration.
+     */
+    public void edgeDrag(String dragFrom, VerticalDirection direction, long dragDuration, TimeUnit timeunit) {
```
damithc: The name is not that intuitive.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -42,35 +45,133 @@ public void editPerson() {
...
+        personListPanel.clickOnPerson(td.alice);
+        EditPersonDialogHandle editPersonDialog =  personListPanel.clickEdit();
+        editPersonDialog.enterFirstName("Peter");
+        editPersonDialog.enterLastName("");
+        editPersonDialog.clickOk();
+
+        assertTrue(editPersonDialog.isInputValidationErrorDialogShown());
+        editPersonDialog.dissmissErrorMessage("Invalid Fields");
+        assertFalse(editPersonDialog.isInputValidationErrorDialogShown());
+    }
```
damithc: May be add a TODO for testing edits during grace period? Can be done in a separate PR.

damithc:
Some cosmetic changes requested.

src/test/java/guitests/guihandles/PersonListPanelHandle.java
yl-coder: cornerDrag?

yl-coder:
Anymore comments before i merge this?

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -42,35 +45,118 @@ public void editPerson() {
...
-        //TODO: confirm the right card is selected after the edit
-        //TODO: confirm status bar is updated correctly
 
+        //Confirm again after the next sync
+        sleep(getTestingConfig().getUpdateInterval(), TimeUnit.MILLISECONDS);
+        assertEquals(newAlice.toString(), personListPanel.getSelectedPerson().toString());
+
+        //Confirm other cards are unaffected.
+        personListPanel.clickOnListView();
+        assertTrue(personListPanel.isListMatching(1, td.benson, td.charlie, td.dan, td.elizabeth));
```
damithc: Minor thing. Create a variant of the method `isListMatching(List<Person>)` that checks for an exact match. The current one checks for sub-list match. The panel can have additional cards at the bottom and the test will still pass.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -42,35 +45,118 @@ public void editPerson() {
...
+
+        //Confirm status bar is updated correctly
+        assertEquals(statusBar.getText(), "Edit Person [ Alice Brown -> Alicia Brownstone ] completed successfully.");
+    }
+
+
+    @Test
```
damithc: Only one blank line between methods.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -42,35 +45,118 @@ public void editPerson() {
...
+        assertEquals(alicePersonCard, td.alice);
+        assertEquals(statusBar.getText(), "Edit Person [ Alice Brown -> Alicia Brownstone ] was cancelled.");
+
+        //New
+        EditPersonDialogHandle addPersonDialog = personListPanel.clickNew();
+        Person pandaWong = new PersonBuilder("Panda", "Wong")
+                .withStreet("Chengdu Panda Street").withCity("Chengdu").withPostalCode("PANDA")
+                .withBirthday("01.01.1979").withGithubUsername("panda").withTags(td.colleagues, td.friends).build();
+        addPersonDialog.enterNewValues(pandaWong);
+        addPersonDialog.clickOk();
```
damithc: Can tweak the GuiHandle to enable chaining in some places. e.g `addPersonDialog.enterNewValues(pandaWong).clickOk()`

damithc: Can be done as a separate PR

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -42,35 +45,118 @@ public void editPerson() {
...
+
+        mainGui.focusOnMainApp();
+        personListPanel.clickOnListView();
+        personListPanel.use_LIST_GOTO_BOTTOM_SEQUENCE();
```
damithc: These three lines seem to have a single purpose i.e. go to bottom. Abstract it.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -42,35 +45,118 @@ public void editPerson() {
...
+        personListPanel.clickOnPerson(pandaWong);
+        PersonCardHandle pandaWongCardHandle = personListPanel.getSelectedCards().get(0);
```
damithc: This line is showing low level details i.e. `get(0)`.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -42,35 +45,118 @@ public void editPerson() {
...
+        assertNotNull(pandaWongCardHandle);
+        assertEquals(pandaWongCardHandle, pandaWong);
+        personListPanel.use_PERSON_CHANGE_CANCEL_ACCELERATOR();
+        assertNull(personListPanel.getPersonCardHandle(pandaWong));
+    }
```
damithc: Do you have a test case to confirm Ctrl+Z does not have any effect after the grace period is over? If not, add a todo or add such a test.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -42,35 +45,118 @@ public void editPerson() {
...
+
+        //Get a reference to the card displaying Alice's details
+        PersonCardHandle alicePersonCard = personListPanel.selectCard(td.alice);
+
+        //Edit Alice to change to new values
+        EditPersonDialogHandle editPersonDialog = personListPanel.use_PERSON_EDIT_ACCELERATOR();
+        editPersonDialog.enterNewValues(newAlice);
+        editPersonDialog.pressEnter();
+
+        mainGui.focusOnMainApp();
```
damithc: This is a low level detail. Consider hiding inside `clickOnPerson()` method.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -42,35 +45,118 @@ public void editPerson() {
...
+        personListPanel.clickOnPerson(newAlice);
+
+        assertNotEquals(alicePersonCard, td.alice);
+        assertEquals(alicePersonCard, newAlice);
```
damithc: Do you need both these?

damithc:
Added some more comments. The test case language is still too verbose at some places, making it hard to understand what is being tested using what steps. Write code at the level of 'intent', not 'mechanism'.
e.g.
**GOOD:**
```
go to bottom of list
click on item
delete using accelerator
ensure grace period is shown
wait till grace period is over
...
```
**VERBOSE:**
```
Click on main window
Click on list
Use go to bottom key sequence
Click on last item
Ensure last item is clicked
...
Use delete accelerator
Sleep 1 sec
Check grace period text
Sleep 3 secs
...
```

damithc:
BTW, it's ok if you merge this first and refine it further as a separate PR. Otherwise @m133225 can't start writing tests.

src/test/java/guitests/PersonEditGuiTest.java
yl-coder: abstracted focusOnMainApp into clickOk. clickOnListView is to ensure shortcut keys work properly. If we will to abstract this one, we will need to abstract it into every shortcut key methods. So i guess leaving it alone will be the best option. As shortcut keys maybe executed consecutively. Having clickOnListView on every shortcut key is going to slow everything down.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -42,35 +45,116 @@ public void editPerson() {
...
+
+        //New
+        EditPersonDialogHandle addPersonDialog = personListPanel.clickNew();
+        Person pandaWong = new PersonBuilder("Panda", "Wong")
+                .withStreet("Chengdu Panda Street").withCity("Chengdu").withPostalCode("PANDA")
+                .withBirthday("01.01.1979").withGithubUsername("panda").withTags(td.colleagues, td.friends).build();
+        addPersonDialog.enterNewValues(pandaWong);
+        addPersonDialog.clickOk();
+
+        personListPanel.clickOnListView(); // To ensure shortcut keys work properly.
```
damithc: From what I remember, this is needed for travis headful mode only. It is a workaround and hence should be hidden (inside GuiRobot, GuiHandle, or GuiTestBase) and automatically executed for shortcuts. We can't expect anybody writing GuiTests to be able to remember when and where to use it.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -42,35 +45,116 @@ public void editPerson() {
...
+
+        personListPanel.use_LIST_GOTO_BOTTOM_SEQUENCE();
+        PersonCardHandle pandaWongCardHandle = personListPanel.getPersonCardHandle(pandaWong);
+        assertEquals(pandaWongCardHandle, pandaWong);
+        personListPanel.use_PERSON_CHANGE_CANCEL_ACCELERATOR();
+        assertNull(personListPanel.getPersonCardHandle(pandaWong));
+
+        //Ensure cancel operation has no side effects after grace period.
```
damithc: This is not what I meant.
You should perform an operation, wait till the grace period is over, and then press the cancel accelerator. We expect the accelerator not to cancel the operation.

src/test/java/guitests/PersonEditGuiTest.java
yl-coder: Clicking on the listview everytime before the shortcuts aren't appropriate for this case. I will think of a better place to hide it.

src/test/java/guitests/PersonEditGuiTest.java
damithc: Use a conditional click? `if(focus not in list), click on list`

src/test/java/guitests/PersonEditGuiTest.java
damithc: Use a conditional click? `if(focus not in list), click on list`

yl-coder: I think the problem was not on the focus. The listview was in focus before the failure. Clicking it just magically make it to work. I give it a shot.

PR #513, opened 2016-07-15
Extract components of UpdateManager into a separate jar #512

### LINKED ISSUE #512: Extract components of UpdateManager into a separate jar
As part of the installer/updater architecture refactoring.

m133225:
Fixes #512.


config/checkstyle/checkstyle.xml
```diff
@@ -89,7 +89,7 @@
       <!-- Validates identifiers for package names against the
         supplied expression. -->
       <!-- Here the default checkstyle rule restricts package name parts to
-        seven characters, this is not in line with common practice at Google.
+        seven characters, this is not in line with commons practice at Google.
```
damithc: Mistake

src/main/java/address/updater/UpdateProgressNotifier.java
```diff
@@ -0,0 +1,55 @@
+package address.updater;
```
damithc: Shouldn't this be 'hubturbo.updater`?  (i.e. not inside addressbook code)

damithc: Never mind. I got it.

m133225: On second thought, this should be moved to commons since it is used by 2 components. Otherwise, it will be difficult to track which components need to be updated (since their source code are placed everywhere).

PR #516, opened 2016-07-17
Extract Launcher into a separate jar #515

### LINKED ISSUE #515: Extract Launcher into a separate jar
- To launch the application with assertion and to allow the user to delete the installer
- Replaced by a newer version through the main application upon update

m133225:
Fixes #515


m133225:
Will also attempt to configure transitive local dependencies. i.e. updateMigrator uses commons, anything that uses updateMigrator should be able to access commons. Currently this is not configured since local dependencies are not located in a maven repo.

PR #537, opened 2016-07-20
Upgrader should keep retrying when trying to replace updater #536

### LINKED ISSUE #536: Upgrader should keep retrying when trying to replace updater
Currently update fails on Windows since it is attempting to replace the updater before it closes.

m133225:
Fixes #536


m133225:
I don't think it's possible for the main application to upgrade the updater jar.
1) If updater jar is used as a run-time dependency: we cannot modify the updater jar
2) If updater jar is used as a separate application: we cannot pass it the notifier object for it to report its status

I think we require 2 components, each with its own update functionalities, in order to able to update any components as desired.

damithc:
The rule we did not consider in the new architecture is that `a jar cannot modify another jar on which  it has a runtime dependency`?

m133225:
Yes, that's probably the case.

damithc:
The solution is for the updatee to launch a separate updator app that will do the update when the updatee is no longer active?

m133225:
Yes, and the updater will be left alone after the update.

PR #545, opened 2016-07-22
Organize existing events #344

### LINKED ISSUE #344: Organize existing events
There is a long list of events in the events package. Maybe we should group them up into different sub-packages for clarity?

m133225:
Fixes #344


damithc:
We are categorizing events based on their origin? In that case, why not move events to packages that raise them? What if there is an event that can be raised by multiple components?

m133225:
I think in general, maybe categorizing based on their origin would be sufficient. Cases where an event can be raised by multiple components likely implies that it is some sort of request to the handler e.g. `SaveDataRequestEvent`, then these could be grouped by their handler instead.

src/test/java/address/storage/StorageManagerTest.java
```diff
@@ -1,6 +1,11 @@
 package address.storage;
 
 import address.events.*;
```
damithc: This line may be redundant.

PR #546, opened 2016-07-22
Fix cloud simulator not using timestamps from file data #541

### LINKED ISSUE #541: Fix cloud simulator not using timestamps from file data
Causes #533

YijinL:
Fixes #541

### DIFF
--- a/src/main/java/address/sync/cloud/CloudSimulator.java
+++ b/src/main/java/address/sync/cloud/CloudSimulator.java
@@ -445,7 +445,7 @@ public class CloudSimulator implements IRemote {
 
     private List<CloudPerson> filterPersonsByTime(List<CloudPerson> personList, LocalDateTime time) {
         return personList.stream()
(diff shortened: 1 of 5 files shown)


src/main/java/address/sync/cloud/CloudSimulator.java
```diff
@@ -445,7 +445,7 @@ private RemoteResponse getEmptyResponse(int responseCode) {
 
     private List<CloudPerson> filterPersonsByTime(List<CloudPerson> personList, LocalDateTime time) {
         return personList.stream()
-                .filter(person -> !person.getLastUpdatedAt().isBefore(time))
+                .filter(person -> person.getLastUpdatedAt().isAfter(time))
```
m133225: I don't think this should be changed. This is based on the `since` [parameter](https://developer.github.com/v3/issues/#parameters) of GitHub Issue API where it will return anything **updated at** or after the time.

PR #547, opened 2016-07-23
Renamed PersonOverview to PersonListPanel

damithc:
I think the name `PersonOverview` came from the original example app, but not suitable for the current UI.


damithc:
Going to merge. @m133225 @yl-coder have a look and comment if any issues.

PR #549, opened 2016-07-24
Restructured ui

damithc:
Experimental restructuring of UI.
Objectives:
- Create a class structure that reflects the UI structure
- Avoid using `Controller` to refer to the class that represents a part of the UI 
- Avoid very long classes
- To extract out loading fxml files into a separate class `UiPartLoader`

image

Just wanted to see if this is viable.
I think I broke a few tests in the process (had to move fast).


damithc:
'Jump to nth card' shortcut triggers an NPE although the jump still happens.

m133225:
Would it be better if the `Dialog` contains the loading/configurations which loads the GUI and handles a `Controller`? i.e. separate the restructured dialog into `loader` and `controller`.

damithc:
Good point. That's the first design I tried i.e. each component had three parts. 
```
[*UiPart]---->[*View] {this is the loader]
  |
  ----------->[*Controller]
```
But after implementing it (this was how I spent my entire Saturday!) I realized 
* all loaders were quite similar and also very small.
* `UiPart` was merely holding the other two objects and not doing anything else
* Most of the code is still in the `Controller`

Then I tried collapsing all three into one class and naming it `*UiPart` instead of `*Controller` but it turns out a controller cannot load itself. That's how I ended up with the separate loader.

src/main/java/address/controller/MainWindow.java
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    /**
+     * Shows the person list panel inside the main Window.
+     */
+    public PersonListPanel createPersonListPanel() {
+        logger.debug("Loading person list panel.");
+        PersonListPanel personListPanel =
+                UiPartLoader.loadUiPart(primaryStage, getPersonListPlaceholder(), new PersonListPanel());
+        personListPanel.configure(ui, modelManager, modelManager.getAllViewablePersonsReadOnly());
```
damithc: It would be even neater if we can combine the above two lines to a static method like this:
```java
PersonListPanel personListPanel = PersonListPanel.load(primaryStage, getPersonListPlaceholder(), 
                                       new PersonListPanel(), ui, modelManager, 
                                       modelManager.getAllViewablePersonsReadOnly());
```
That way, we can move the entire construction logic inside the respective UiPart class.

m133225:
Instead of the UI part having reference to both objects, I was thinking more along this line:
```
[*UiPart] ----> [*Dialog] ----> [*Controller]
```
The dialog here thus:
- Contains constants such as the default size of the dialog
- Keeps and returns results if there are (e.g. a `Tag`)
- Extends a `BaseDialog` class/implements interface which contains a `load()`/`configure()`
- Handles any request for the resizing of the window

This might be a better, and we can abstract away methods such as `setEscKeyToDismiss` into the base class.

src/main/java/address/ui/Ui.java
```diff
@@ -1,39 +1,445 @@
...
+        EventManager.getInstance().registerHandler(this);
     }
 
-    public void showAlertDialogAndWait(Alert.AlertType alertType, String alertTitle, String headerText, String contentText) {
-        mainController.showAlertDialogAndWait(alertType, alertTitle, headerText, contentText);
+    public void start(Stage primaryStage) {
+        try {
+            logger.info("Starting main UI.");
+
+            mainWindow = MainWindow.load(primaryStage, config, prefs, mainApp, this, modelManager);
```
damithc: Moved the construction logic completely from the call site and put it in the respective class. This is an example of the clean call site.

PR #555, opened 2016-07-25
Cleanup code base

m133225:
- Removed unused variables
- Check code styles/access modifiers
- Run checkstyle
etc


m133225:
Will merge this after #549 since it will cause quite abit of conflicts.

PR #557, opened 2016-07-25
Improvement and more gui test cases.

### LINKED ISSUE #562: Add full GUI tests for adding a person
as per title

### LINKED ISSUE #565: PersonCardHandle::isShowingGracePeriod fails sometimes
Seems to fail when performing edits on a person other than alice.
- `isPendingStateRootVisible()` returns `false` instead of the required `true`

yl-coder:
As per title.
Fixes #565
Fixes #562

### DIFF
--- a/src/test/java/address/testutil/TypicalTestData.java
+++ b/src/test/java/address/testutil/TypicalTestData.java
@@ -23,10 +23,21 @@ public class TypicalTestData {
      * And so on.
      */
     public Person alice = new Person("Alice", "Brown", 1);
+    public Person aliceEdited = new PersonBuilder(alice.copy()).withFirstName("Alicia").withLastName("Brownstone")
+                                                               .withStreet("81th Wall Street").withCity("New York")
+                                                               .withPostalCode("41452").withBirthday("11.09.1983")
+                                                               .withGithubUsername("alicia").withTags(friends).build();
     public Person benson = new Person("Benson", "Christopher Dean", 2);
+    public Person bensonEdited = new PersonBuilder(benson.copy()).withFirstName("Ben").withLastName("Chris")
+                                                                 .withStreet("Pittsburgh Square").withCity("Pittsburg")
+                                                                 .withPostalCode("42445").withGithubUsername("ben").build();
     public Person charlie = new Person("Charlie", "Davidson", 3);
+    public Person charlieEdited = new PersonBuilder(charlie.copy()).withFirstName("Charlotte").withCity("Texas")
+                                                                   .withGithubUsername("charlotte").build();
     public Person dan = new Person("Dan", "Edwards", 4);
+    public Person danEdited = new PersonBuilder(dan.copy()).withLastName("Edmonton").withBirthday("03.01.1995").build();
     public Person elizabeth = new Person("Elizabeth", "F. Green", 5);
+    public Person elizabethEdited = new PersonBuilder(elizabeth.copy()).withLastName("Green").build();
     public Person fiona = new PersonBuilder("Fiona", "Wong", 6).withStreet("51th street").withCity("New York")
                                                                .withPostalCode("51245").withBirthday("01.01.1980")
                                                                .withGithubUsername("fiona").withTags(friends).build();
--- a/src/test/java/address/testutil/TypicalTestData.java
+++ b/src/test/java/address/testutil/TypicalTestData.java
@@ -4,12 +4,19 @@ import address.model.datatypes.AddressBook;
 import address.model.datatypes.person.Person;
 import address.model.datatypes.tag.Tag;
 
+import java.util.ArrayList;
+import java.util.List;
+
 /**
  * Typical data set used for testing
  */
 public class TypicalTestData {
 
+    public Tag colleagues = new Tag("colleagues");
+    public Tag friends = new Tag("friends");
+
     /*
+     * TODO: add more details to these persons.
      * Note that staring letter of names have some pattern.
      * First person's name start with A and her second name starts with B.
      * The Second person's name starts with B and his other names start with C and D.
@@ -20,13 +27,18 @@ public class TypicalTestData {
     public Person charlie = new Person("Charlie", "Davidson", 3);
     public Person dan = new Person("Dan", "Edwards", 4);
     public Person elizabeth = new Person("Elizabeth", "F. Green", 5);
-    //TODO: add more details to these persons.
+    public Person fiona = new PersonBuilder("Fiona", "Wong", 6).withStreet("51th street").withCity("New York")
+                                                               .withPostalCode("51245").withBirthday("01.01.1980")
+                                                               .withGithubUsername("fiona").withTags(friends).build();
+    public Person george = new PersonBuilder("George", "David", 7).withStreet("8th Ave").withCity("Chicago")
+                                                                  .withPostalCode("25614").withBirthday("01.12.1988")
+                                                                  .withGithubUsername("george").withTags(friends, colleagues)
+                                                                  .build();
 
-    public Tag colleagues = new Tag("colleagues");
-    public Tag friends = new Tag("friends");
+
 
     public AddressBook book;
-
+    public List<Person> samples;
 
     public TypicalTestData() {
         book = new AddressBook();
@@ -38,5 +50,7 @@ public class TypicalTestData {
         book.addTag(colleagues);
         book.addTag(friends);
     }
+
+
 
 }
--- a/src/test/java/guitests/KeyBindingsGuiTest.java
+++ b/src/test/java/guitests/KeyBindingsGuiTest.java
@@ -14,8 +14,11 @@ import static org.junit.Assert.*;
 
 /**
  * Tests key bindings through the GUI
- * Keybindings tests will fail in travis ci headfull
- * One magic solution is to click on the listview first before firing shortcuts.
+ * Special Note:
+ * In travis ci headlfull testing, the following were observed:
+ * 1) Keybinding tests failed as shortcuts are not triggered.
+ *    - First shortcuts get triggered as normal, second shortcuts onwards failed to get triggered.
+ * Solution: Click on the listview first before firing shortcuts. (Use personListPanel.clickOnListView())
  */
 public class KeyBindingsGuiTest extends GuiTestBase {
 
--- a/src/test/java/guitests/KeyBindingsGuiTest.java
+++ b/src/test/java/guitests/KeyBindingsGuiTest.java
@@ -29,7 +29,7 @@ public class KeyBindingsGuiTest extends GuiTestBase {
         personListPanel.use_LIST_ENTER_SHORTCUT();
         assertTrue(personListPanel.isSelected(td.alice));
 
-        personListPanel.clickOnListView();
+        //personListPanel.clickOnListView();
 
         personListPanel.use_LIST_JUMP_TO_INDEX_SHORTCUT(4);
         assertTrue(personListPanel.isSelected(td.dan));
@@ -72,8 +72,7 @@ public class KeyBindingsGuiTest extends GuiTestBase {
         //Focus on MainApp, could not abstract into tagPersonDialog.close(), as closing tagDialog may lead to
         //focusing on EditDialog or MainApp.
         mainGui.focusOnMainApp();
-
-        personListPanel.clickOnListView();
+        //personListPanel.clickOnListView();
 
         //======== others ============================
         personListPanel.use_LIST_JUMP_TO_INDEX_SHORTCUT(2);
@@ -93,7 +92,7 @@ public class KeyBindingsGuiTest extends GuiTestBase {
      */
     @Test
     public void testHotKeys() {
-        personListPanel.clickOnListView();
+        //personListPanel.clickOnListView();
 
         mainGui.use_APP_MINIMIZE_HOTKEY();
         assertTrue(mainGui.isMinimized());
--- a/src/test/java/guitests/KeyBindingsGuiTest.java
+++ b/src/test/java/guitests/KeyBindingsGuiTest.java
@@ -14,6 +14,8 @@ import static org.junit.Assert.*;
 
 /**
  * Tests key bindings through the GUI
+ * Keybindings tests will fail in travis ci headfull
+ * One magic solution is to click on the listview first before firing shortcuts.
  */
 public class KeyBindingsGuiTest extends GuiTestBase {
 
@@ -28,8 +30,6 @@ public class KeyBindingsGuiTest extends GuiTestBase {
 
         personListPanel.use_LIST_ENTER_SHORTCUT();
         assertTrue(personListPanel.isSelected(td.alice));
-
-        //personListPanel.clickOnListView();
 
         personListPanel.use_LIST_JUMP_TO_INDEX_SHORTCUT(4);
         assertTrue(personListPanel.isSelected(td.dan));
@@ -72,7 +72,6 @@ public class KeyBindingsGuiTest extends GuiTestBase {
         //Focus on MainApp, could not abstract into tagPersonDialog.close(), as closing tagDialog may lead to
         //focusing on EditDialog or MainApp.
         mainGui.focusOnMainApp();
-        //personListPanel.clickOnListView();
 
         //======== others ============================
         personListPanel.use_LIST_JUMP_TO_INDEX_SHORTCUT(2);
@@ -92,7 +91,6 @@ public class KeyBindingsGuiTest extends GuiTestBase {
      */
     @Test
     public void testHotKeys() {
-        //personListPanel.clickOnListView();
 
         mainGui.use_APP_MINIMIZE_HOTKEY();
         assertTrue(mainGui.isMinimized());
--- a/src/test/java/guitests/KeyBindingsGuiTest.java
+++ b/src/test/java/guitests/KeyBindingsGuiTest.java
@@ -71,12 +71,12 @@ public class KeyBindingsGuiTest extends GuiTestBase {
         sleepForGracePeriod();
         assertTrue(personListPanel.contains(td.charlie)); // still in the list even after grace period
 
+        EditPersonDialogHandle editPersonDialogHandle = personListPanel.use_PERSON_EDIT_ACCELERATOR();
+        assertTrue(editPersonDialogHandle.isShowingPerson(td.charlie));
+        editPersonDialogHandle.clickCancel();
+
         TagPersonDialogHandle tagPersonDialog = personListPanel.use_PERSON_TAG_ACCELERATOR();
         tagPersonDialog.close();
-
-        //Focus on MainApp, could not abstract into tagPersonDialog.close(), as closing tagDialog may lead to
-        //focusing on EditDialog or MainApp.
-        mainGui.focusOnMainApp();
 
         //======== others ============================
         personListPanel.use_LIST_JUMP_TO_INDEX_SHORTCUT(2);
--- a/src/test/java/guitests/KeyBindingsGuiTest.java
+++ b/src/test/java/guitests/KeyBindingsGuiTest.java
@@ -71,10 +71,6 @@ public class KeyBindingsGuiTest extends GuiTestBase {
         sleepForGracePeriod();
         assertTrue(personListPanel.contains(td.charlie)); // still in the list even after grace period
 
-        EditPersonDialogHandle editPersonDialogHandle = personListPanel.use_PERSON_EDIT_ACCELERATOR();
-        assertTrue(editPersonDialogHandle.isShowingPerson(td.charlie));
-        editPersonDialogHandle.clickCancel();
-
         TagPersonDialogHandle tagPersonDialog = personListPanel.use_PERSON_TAG_ACCELERATOR();
         tagPersonDialog.close();
 
--- a/src/test/java/guitests/KeyBindingsGuiTest.java
+++ b/src/test/java/guitests/KeyBindingsGuiTest.java
@@ -18,7 +18,9 @@ import static org.junit.Assert.*;
  * In travis ci headlfull testing, the following were observed:
  * 1) Keybinding tests failed as shortcuts are not triggered.
  *    - First shortcuts get triggered as normal, second shortcuts onwards failed to get triggered.
- * Solution: Click on the listview first before firing shortcuts. (Use personListPanel.clickOnListView())
+ * Solution(Not implemented, as travis headfull testing is not needed currently):
+ * - Click on the listview first before firing shortcuts. (Use personListPanel.clickOnListView())
+ *   (Seems to work, but no concrete reason on why it works).
  */
 public class KeyBindingsGuiTest extends GuiTestBase {
 
--- a/src/test/java/guitests/PersonDeleteGuiTest.java
+++ b/src/test/java/guitests/PersonDeleteGuiTest.java
@@ -20,7 +20,6 @@ public class PersonDeleteGuiTest extends GuiTestBase {
     @Test
     public void cancelOperation_usingAccelerator() {
 
-        //Delete
         PersonCardHandle aliceCard = personListPanel.selectCard(td.alice);
         personListPanel.use_PERSON_DELETE_ACCELERATOR();
         assertTrue(aliceCard.isShowingGracePeriod("Deleting"));
--- a/src/test/java/guitests/PersonDeleteGuiTest.java
+++ b/src/test/java/guitests/PersonDeleteGuiTest.java
@@ -1,0 +1,38 @@
+package guitests;
+
+import address.model.datatypes.AddressBook;
+import address.model.datatypes.person.Person;
+import address.testutil.PersonBuilder;
+import guitests.guihandles.EditPersonDialogHandle;
+import guitests.guihandles.PersonCardHandle;
+import org.junit.Test;
+
+import static junit.framework.TestCase.assertFalse;
+import static org.junit.Assert.*;
+
+/**
+ * System tests for 'Delete person' feature.
+ */
+public class PersonDeleteGuiTest extends GuiTestBase {
+
+    @Override
+    protected AddressBook getInitialData() {
+        return td.book;
+    }
+
+    @Test
+    public void cancelOperation_usingAccelerator() {
+
+        //Delete
+        PersonCardHandle aliceCard = personListPanel.selectCard(td.alice);
+        personListPanel.use_PERSON_DELETE_ACCELERATOR();
+        assertTrue(aliceCard.isShowingGracePeriod("Deleting"));
+
+        personListPanel.use_PERSON_CHANGE_CANCEL_ACCELERATOR();
+        assertFalse(aliceCard.isShowingGracePeriod("Deleting"));
+        assertEquals(statusBar.getText(), "Delete Person [ " + aliceCard.getFirstName() + " "
+                     + aliceCard.getLastName() + " ] was cancelled.");
+
+
+    }
+}
--- a/src/test/java/guitests/PersonDeleteGuiTest.java
+++ b/src/test/java/guitests/PersonDeleteGuiTest.java
@@ -1,9 +1,6 @@
 package guitests;
 
 import address.model.datatypes.AddressBook;
-import address.model.datatypes.person.Person;
-import address.testutil.PersonBuilder;
-import guitests.guihandles.EditPersonDialogHandle;
 import guitests.guihandles.PersonCardHandle;
 import org.junit.Test;
 
@@ -32,7 +29,5 @@ public class PersonDeleteGuiTest extends GuiTestBase {
         assertFalse(aliceCard.isShowingGracePeriod("Deleting"));
         assertEquals(statusBar.getText(), "Delete Person [ " + aliceCard.getFirstName() + " "
                      + aliceCard.getLastName() + " ] was cancelled.");
-
-
     }
 }
--- a/src/test/java/guitests/PersonDeleteGuiTest.java
+++ b/src/test/java/guitests/PersonDeleteGuiTest.java
@@ -1,13 +1,17 @@
 package guitests;
 
 import address.model.datatypes.AddressBook;
+import guitests.guihandles.HeaderStatusBarHandle;
 import guitests.guihandles.PersonCardHandle;
 import org.junit.Test;
+
+import java.util.Optional;
 
 import static junit.framework.TestCase.assertFalse;
 import static org.junit.Assert.*;
 
 /**
+ * TODO: Finish the full use cases of delete person
  * System tests for 'Delete person' feature.
  */
 public class PersonDeleteGuiTest extends GuiTestBase {
@@ -26,7 +30,7 @@ public class PersonDeleteGuiTest extends GuiTestBase {
 
         personListPanel.use_PERSON_CHANGE_CANCEL_ACCELERATOR();
         assertFalse(aliceCard.isShowingGracePeriod("Deleting"));
-        assertEquals(statusBar.getText(), "Delete Person [ " + aliceCard.getFirstName() + " "
-                     + aliceCard.getLastName() + " ] was cancelled.");
+        assertEquals(HeaderStatusBarHandle.formatCancelledMessage(HeaderStatusBarHandle.Type.DELETE,
+                     td.alice.fullName(), Optional.empty()), statusBar.getText());
     }
 }
--- a/src/test/java/guitests/PersonEditGuiTest.java
+++ b/src/test/java/guitests/PersonEditGuiTest.java
@@ -127,11 +127,12 @@ public class PersonEditGuiTest extends GuiTestBase {
         //Edit Alice to change to new values
         EditPersonDialogHandle editPersonDialog = personListPanel.use_PERSON_EDIT_ACCELERATOR();
         editPersonDialog.enterNewValues(newAlice).pressEnter();
-
+        assertTrue(alicePersonCard.isShowingGracePeriod("Editing"));
         personListPanel.clickOnPerson(newAlice);
         assertEquals(alicePersonCard, newAlice);
         personListPanel.use_PERSON_CHANGE_CANCEL_ACCELERATOR();
         assertEquals(alicePersonCard, td.alice);
+        assertFalse(alicePersonCard.isShowingGracePeriod("Editing"));
         assertEquals(statusBar.getText(), "Edit Person [ Alice Brown -> Alicia Brownstone ] was cancelled.");
 
         //New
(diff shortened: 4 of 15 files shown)


src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -77,7 +74,7 @@ public void editPerson_usingAccelerator() {
     public void editPerson_usingContextMenu() {
         personListPanel.rightClickOnPerson(td.alice);
         EditPersonDialogHandle editPersonDialog = personListPanel.clickOnContextMenu(
-                PersonListPanelHandle.ContextMenuChoice.EDIT);
+                                                                          PersonListPanelHandle.ContextMenuChoice.EDIT);
```
damithc: Try breaking after `=` too.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -139,22 +135,57 @@ public void cancelOperation_usingAccelerator() {
...
+        Person pandaWong = new PersonBuilder("Panda", "Wong")
+                .withStreet("Chengdu Panda Street").withCity("Chengdu").withPostalCode("PANDA")
+                .withBirthday("01.01.1979").withGithubUsername("panda").withTags(td.colleagues, td.friends).build();
+        addPersonDialog.enterNewValues(pandaWong).clickOk();
+        sleepForGracePeriod();
+        personListPanel.clickOnListView();
+        personListPanel.use_LIST_GOTO_BOTTOM_SEQUENCE();
+        personListPanel.use_PERSON_CHANGE_CANCEL_ACCELERATOR();
+        assertEquals(personListPanel.getPersonCardHandle(pandaWong), pandaWong);
+    }
```
damithc: To make testing efficient, you can test this scenario in another test that already has a completed edit op.
Inside this method simply have a comment explaining in which test this functionality is tested.

src/test/java/guitests/guihandles/PersonListPanelHandle.java
```diff
@@ -76,10 +76,13 @@ public ReadOnlyViewablePerson getSelectedPerson() {
     }
 
     /**
-     * Clicks on the middle of the Listview.
+     * Clicks on the middle of the Listview if the listview is not in focus
      * In order for headfull testing to work in travis ci, listview needs to be clicked before firing hot keys.
      */
     public void clickOnListView() {
```
damithc: If you go ahead with this, rename method to 'ensureListViewHasFocus' (because clicking is no longer guaranteed).

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -52,21 +51,20 @@ public void editPerson_usingAccelerator() {
         assertEquals(alicePersonCard, newAlice);
         assertTrue(personListPanel.isSelected(newAlice));
 
-
         //Confirm right values are displayed after grace period is over
         sleepForGracePeriod();
         assertEquals(alicePersonCard, newAlice);
 
+        //Confirm cancel operation does not cancel the edit after the grace period.
+        personListPanel.use_PERSON_CHANGE_CANCEL_ACCELERATOR();
```
damithc: This is not completely correct. The cancel is inserted between two asserts with two different purposes. By right, you should do both asserts first, cancel, and repeat both asserts again.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -139,22 +136,45 @@ public void cancelOperation_usingAccelerator() {
         Person pandaWong = new PersonBuilder("Panda", "Wong")
                 .withStreet("Chengdu Panda Street").withCity("Chengdu").withPostalCode("PANDA")
                 .withBirthday("01.01.1979").withGithubUsername("panda").withTags(td.colleagues, td.friends).build();
-        addPersonDialog.enterNewValues(pandaWong);
-        addPersonDialog.clickOk();
-
-        personListPanel.clickOnListView(); // To ensure shortcut keys work properly.
+        addPersonDialog.enterNewValues(pandaWong).clickOk();
 
         personListPanel.use_LIST_GOTO_BOTTOM_SEQUENCE();
```
damithc: Why is there is a need to go to the bottom here? If you simply want to go to a specific card, we should not use go_to_bottom as it is misleading to the reader.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -139,22 +136,45 @@ public void cancelOperation_usingAccelerator() {
...
+                .withBirthday("01.01.1979").withGithubUsername("alicebrown123").withTags(td.colleagues, td.friends).build();
 
-        //Ensure cancel operation has no side effects after grace period.
+        //Get a reference to the card displaying Alice's details
+        PersonCardHandle alicePersonCard = personListPanel.selectCard(td.alice);
+
+        //Edit Alice to change to new values
+        EditPersonDialogHandle editPersonDialog = personListPanel.use_PERSON_EDIT_ACCELERATOR();
+        editPersonDialog.enterNewValues(newAlice).pressEnter();
+
```
damithc: Ensure pending state is displayed?

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -139,22 +136,45 @@ public void cancelOperation_usingAccelerator() {
...
+        //Edit Alice again during pending state.
+        personListPanel.clickOnPerson(newAlice);
+        Person newerAlice = new PersonBuilder(newAlice.copy()).withFirstName("Felicia").withLastName("Yellowstone")
+                .withStreet("street updated").withCity("Malaysia").withPostalCode("321321")
+                .withBirthday("11.11.1979").withGithubUsername("yellowstone").withTags(td.colleagues).build();
+        editPersonDialog = personListPanel.use_PERSON_EDIT_ACCELERATOR();
+        editPersonDialog.enterNewValues(newerAlice).pressEnter();
```
damithc: Ensure pending state display is frozen?

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -139,22 +136,45 @@ public void cancelOperation_usingAccelerator() {
...
+
+        //Ensure card is displaying Felicia before and after grace period.
+        assertEquals(alicePersonCard, newerAlice);
         sleepForGracePeriod();
-        assertNull(personListPanel.getPersonCardHandle(pandaWong));
-        personListPanel.isListMatching(td.alice, td.benson, td.charlie, td.dan, td.elizabeth);
+        assertEquals(alicePersonCard, newerAlice);
```
damithc: Ensure other cards are unaffected?

damithc:
Need to do more comprehensive verification. Any GUI change that is expected should be verified. Any GUI element that is related but not expected to change should be verified to remain same.

src/test/java/guitests/PersonEditGuiTest.java
yl-coder: actually assertEquals(alicePersonCard, newAlice) and  assertEquals(newAlice.toString(), personListPanel.getSelectedPerson().toString()); is the same. I will standardize to the former.

src/test/java/guitests/PersonEditGuiTest.java
yl-coder: Already tested in editPerson_usingAccelerator().

yl-coder: Is there a need to test it again?

src/test/java/guitests/KeyBindingsGuiTest.java
```diff
@@ -14,6 +14,8 @@
 
 /**
  * Tests key bindings through the GUI
+ * Keybindings tests will fail in travis ci headfull
+ * One magic solution is to click on the listview first before firing shortcuts.
```
damithc: This explanation is good enough for those of us who know what happened, but not for a new developer. Comments should be written with new developers in mind, not as 'note to self'.

src/test/java/guitests/PersonEditGuiTest.java
damithc: I thought they are the same, but when I tried it, they gave different results. Not sure what the cause is. It looked like the underlying list changed only after the grace period is over.

src/test/java/guitests/PersonEditGuiTest.java
damithc: Yes, no harm doing again. Sometimes things work the first time but not the second time.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -125,36 +126,70 @@ public void cancelOperation_usingAccelerator() {
...
+        Person newerAlice = new PersonBuilder(newAlice.copy()).withFirstName("Felicia").withLastName("Yellowstone")
+                .withStreet("street updated").withCity("Malaysia").withPostalCode("321321")
+                .withBirthday("11.11.1979").withGithubUsername("yellowstone").withTags(td.colleagues).build();
+
+        editPersonDialog = personListPanel.use_PERSON_EDIT_ACCELERATOR();
+
+        //Ensure grace period is not counting down while editing person.
+        assertTrue(alicePersonCard.isGracePeriodFrozen());
+
+        editPersonDialog.enterNewValues(newerAlice).pressEnter();
```
damithc: Before entering new values, you should verify that the form is showing the pending values from the previous edit.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -125,36 +126,70 @@ public void cancelOperation_usingAccelerator() {
...
+
+        //Ensure card is displaying Felicia before and after grace period.
+        assertEquals(alicePersonCard, newerAlice);
```
damithc: How do you confirm if the grace period restarted from 0? Perhaps the grace period was never activated?

src/test/java/guitests/KeyBindingsGuiTest.java
```diff
@@ -14,6 +14,11 @@
 
 /**
  * Tests key bindings through the GUI
+ * Special Note:
+ * In travis ci headlfull testing, the following were observed:
+ * 1) Keybinding tests failed as shortcuts are not triggered.
+ *    - First shortcuts get triggered as normal, second shortcuts onwards failed to get triggered.
+ * Solution: Click on the listview first before firing shortcuts. (Use personListPanel.clickOnListView())
```
damithc: It's not clear if the solution is being used in the current code or it's just a suggestion. Also mention that it seems to work although we don't know exactly why.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -76,16 +73,20 @@ public void editPerson_usingAccelerator() {
...
         assertTrue(editPersonDialog.isValidEditDialog());
+
+        //Rest of the edit process tested in editPerson_usingAccelerator
     }
 
     @Test
     public void editPerson_usingEditButton() {
         personListPanel.clickOnPerson(td.alice);
         EditPersonDialogHandle editPersonDialog =  personListPanel.clickEdit();
         assertTrue(editPersonDialog.isValidEditDialog());
```
damithc: The test is weak. It should check if it shows the edit dialog for the correct person.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -125,36 +126,73 @@ public void cancelOperation_usingAccelerator() {
 
         //Edit Alice to change to new values
```
damithc: I thought this test is only for edits, not deletes?

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -125,36 +126,73 @@ public void cancelOperation_usingAccelerator() {
...
         assertEquals(statusBar.getText(), "Edit Person [ Alice Brown -> Alicia Brownstone ] was cancelled.");
 
         //New
         EditPersonDialogHandle addPersonDialog = personListPanel.clickNew();
         Person pandaWong = new PersonBuilder("Panda", "Wong")
                 .withStreet("Chengdu Panda Street").withCity("Chengdu").withPostalCode("PANDA")
                 .withBirthday("01.01.1979").withGithubUsername("panda").withTags(td.colleagues, td.friends).build();
-        addPersonDialog.enterNewValues(pandaWong);
-        addPersonDialog.clickOk();
+        addPersonDialog.enterNewValues(pandaWong).clickOk();
```
damithc: Same here. I prefer to have Edit, Delete, and New in three separate tests. When we try to test too many things at once, some subtle test cases can get left out.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -125,36 +126,73 @@ public void cancelOperation_usingAccelerator() {
...
 
-        personListPanel.clickOnListView(); // To ensure shortcut keys work properly.
-
-        personListPanel.use_LIST_GOTO_BOTTOM_SEQUENCE();
+        personListPanel.navigateToPerson(pandaWong);
         PersonCardHandle pandaWongCardHandle = personListPanel.getPersonCardHandle(pandaWong);
```
damithc: The `navigate` method can return either the personListPanel (for chaining) or the handle to the card (to avoid an extra call).

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -125,36 +126,73 @@ public void cancelOperation_usingAccelerator() {
...
-        assertEquals(pandaWongCardHandle, pandaWong);
+        assertTrue(pandaWongCardHandle.isShowingGracePeriod("Adding"));
+        assertEquals(pandaWongCardHandle, pandaWong); //Ensure correct state before cancelling.
```
damithc: This doesn't sound right. How come a card is equals to a person?

damithc:
A few more comments added.

src/test/java/guitests/PersonEditGuiTest.java
yl-coder: Right, personListPanel.getSelectedPerson().toString() gives the backing person details. I think it is not appropriate to test it in guitest. Checking if the GUI details match is the one we should test in GUITest.

src/test/java/guitests/PersonEditGuiTest.java
yl-coder: Will moved them into their respective new test classes

src/test/java/guitests/PersonEditGuiTest.java
yl-coder: That is not tested yet. Testing if the grace period actually restarts is abit tricky, may lead to indeterministic failure.

yl-coder:
Abit of refactoring done. Instead of using edit accelerator, created a editPerson(person) method which does the low level testing inside, and hide the implementation details of opening an editDialog.  Moved the edit accelerator into keybindingguitest, as i think it is more appropriate there.

src/test/java/guitests/PersonDeleteGuiTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @Override
+    protected AddressBook getInitialData() {
+        return td.book;
+    }
+
+    @Test
+    public void cancelOperation_usingAccelerator() {
+
+        //Delete
```
damithc: This comment is redundant.

damithc: Sorry, didn't realize this is still ongoing. Let me know when ready for review.

src/test/java/guitests/PersonDeleteGuiTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+        PersonCardHandle aliceCard = personListPanel.selectCard(td.alice);
+        personListPanel.use_PERSON_DELETE_ACCELERATOR();
+        assertTrue(aliceCard.isShowingGracePeriod("Deleting"));
+
+        personListPanel.use_PERSON_CHANGE_CANCEL_ACCELERATOR();
+        assertFalse(aliceCard.isShowingGracePeriod("Deleting"));
+        assertEquals(statusBar.getText(), "Delete Person [ " + aliceCard.getFirstName() + " "
+                     + aliceCard.getLastName() + " ] was cancelled.");
+    }
```
damithc: I assume you'll add more tests to this in a different PR?

damithc: BTW, you may want to hide the status message format inside a method. e.g.
`assertEquals(statusBar.getText(), statusBar.getDeleteCancelledStatus(td.alice))`

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -39,60 +38,56 @@ public void editPerson_usingAccelerator() {
...
         PersonCardHandle alicePersonCard = personListPanel.getPersonCardHandle(td.alice);
 
-        //Edit Alice to change to new values
-        personListPanel.clickOnPerson(td.alice);
-        EditPersonDialogHandle editPersonDialog = personListPanel.use_PERSON_EDIT_ACCELERATOR();
-        editPersonDialog.enterNewValues(newAlice);
-        editPersonDialog.pressEnter();
+        personListPanel.rightClickOnPerson(td.alice);
+        EditPersonDialogHandle editPersonDialog =
+                personListPanel.clickOnContextMenu(PersonListPanelHandle.ContextMenuChoice.EDIT);
```
damithc: It's not correct for context menu click to return a handle to edit dialog. A context menu click may not always lead to an edit op.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -39,60 +38,56 @@ public void editPerson_usingAccelerator() {
...
+        assertTrue(editPersonDialog.isShowingPerson(td.alice));
+        editPersonDialog.enterNewValues(newAlice).pressEnter();
 
         //Confirm pending state correctness
         assertTrue(alicePersonCard.isShowingGracePeriod("Editing"));
 
         //Confirm the right card is selected after the edit
-        assertEquals(alicePersonCard, newAlice);
```
damithc: The comment is relevant for line 52, not 51.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -39,60 +38,56 @@ public void editPerson_usingAccelerator() {
...
 
-    @Test
-    public void editPerson_usingContextMenu() {
-        personListPanel.rightClickOnPerson(td.alice);
-        EditPersonDialogHandle editPersonDialog = personListPanel.clickOnContextMenu(
-                PersonListPanelHandle.ContextMenuChoice.EDIT);
-        assertTrue(editPersonDialog.isValidEditDialog());
     }
 
+    //TODO: This maybe should not be here. A separate class to test the button does what it suppose to do?
```
damithc: It's ok to keep it here. It's the edit button and this is the edit test. Remove the comment.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -39,60 +38,56 @@ public void editPerson_usingAccelerator() {
...
     @Test
     public void editPerson_usingEditButton() {
         personListPanel.clickOnPerson(td.alice);
-        EditPersonDialogHandle editPersonDialog =  personListPanel.clickEdit();
-        assertTrue(editPersonDialog.isValidEditDialog());
+        EditPersonDialogHandle editPersonDialog =  personListPanel.clickEdit();;
+        assertTrue(editPersonDialog.isShowingPerson(td.alice));
```
damithc: Don't always use the first card. A buggy code can return the first card even when a different card is selected. This test will not catch such a bug.

damithc: Use different cards for different tests. Sometimes the first card, sometimes the last, sometimes a middle one. That way, you maximize the chance of catching bug you weren't even targetting.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -104,18 +99,27 @@ public void editPerson_dataValidation() {
 
     @Test
     public void cancelOperation_usingAccelerator() {
+        Person newAlice = new PersonBuilder(td.alice.copy()).withFirstName("Alicia").withLastName("Brownstone")
+                .withStreet("Updated street").withCity("Singapore").withPostalCode("123123")
+                .withBirthday("01.01.1979").withGithubUsername("alicebrown123").withTags(td.colleagues, td.friends).build();
```
damithc: I know you are following my code, but on retrospect, this declaration should be done just before the newAlice is used. In general, declare variables close to where they are used first.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -104,18 +99,27 @@ public void editPerson_dataValidation() {
...
-        personListPanel.use_PERSON_DELETE_ACCELERATOR();
-        assertTrue(aliceCard.isShowingGracePeriod("Deleting"));
+        //Get a reference to the card displaying Alice's details
+        PersonCardHandle alicePersonCard = personListPanel.selectCard(td.alice);
 
+        //Edit Alice to change to new values
+        EditPersonDialogHandle editPersonDialog = personListPanel.editPerson(td.alice);
+        editPersonDialog.enterNewValues(newAlice).pressEnter();
+        assertTrue(alicePersonCard.isShowingGracePeriod("Editing"));
+        personListPanel.clickOnPerson(newAlice);
```
damithc: Why click? Shouldn't card be automatically selected?

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -124,37 +128,35 @@ public void cancelOperation_usingAccelerator() {
...
-        editPersonDialog.enterNewValues(newAlice);
-        editPersonDialog.pressEnter();
+        EditPersonDialogHandle editPersonDialog = personListPanel.editPerson(td.alice);
+        editPersonDialog.enterNewValues(newAlice).pressEnter();
 
+        //Ensure grace period is showing
+        assertTrue(alicePersonCard.isShowingGracePeriod("Editing"));
+
+        //Edit Alice again during pending state.
         personListPanel.clickOnPerson(newAlice);
```
damithc: Same here. click is not needed?

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -124,37 +128,35 @@ public void cancelOperation_usingAccelerator() {
...
-        assertNull(personListPanel.getPersonCardHandle(pandaWong));
+        //Ensure grace period is not counting down while editing person.
+        assertTrue(alicePersonCard.isGracePeriodFrozen());
+
+        editPersonDialog.enterNewValues(newerAlice).pressEnter();
 
-        //Ensure cancel operation has no side effects after grace period.
+        //Ensure card is displaying Felicia before and after grace period.
+        assertMatching(alicePersonCard, newerAlice);
         sleepForGracePeriod();
```
damithc: This test can be stronger. Currently, it does not verify that the grace period restarts the count down, not continue from the previous count. At least add a TODO to acknowledge the missing verification.

src/test/java/guitests/PersonNewGuiTest.java
```diff
@@ -0,0 +1,121 @@
...
+                .withBirthday("01.01.1979").withGithubUsername("panda").withTags(td.colleagues, td.friends).build();
+        addPersonDialog.enterNewValues(pandaWong).clickOk();
+
+        PersonCardHandle pandaWongCard = personListPanel.navigateToPerson(pandaWong);
+        assertTrue(pandaWongCard.isShowingGracePeriod("Adding"));
+        assertMatching(pandaWongCard, pandaWong);
+
+        //Create Person B
+        addPersonDialog = personListPanel.clickNew();
+        Person ipMan = new PersonBuilder("IP", "Man")
```
damithc: Don't get cute with test values. Especially, don't use currently popular buzzwords, as the code is supposed to outlast the latest fads. Stick the boring but meaningful names. For example, our typical test data follow a specific pattern because it helps us detect problems. Alice comes first, and that's why her name starts with A. If we see Alice in the middle of the list, we know something went wrong.

src/test/java/guitests/PersonNewGuiTest.java
```diff
@@ -0,0 +1,121 @@
...
+        assertTrue(personCardHandle.isShowingGracePeriod("Adding"));
+        sleepForGracePeriod();
+        assertNull(personListPanel.contains(pandaWong)); //Make sure Person A doesn't appear again.
+        assertTrue(personListPanel.contains(ipMan));
+    }
+
+    @Test
+    public void addPerson_multipleAdd() {
+
+        //Create Person A
```
damithc: Person A?

src/test/java/guitests/PersonNewGuiTest.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    @Test
+    public void addPerson_cancelDialog() {
+        EditPersonDialogHandle addPersonDialog = personListPanel.clickNew();
+
+        Person pandaWong = new PersonBuilder("Panda", "Wong")
+                .withStreet("Chengdu Panda Street").withCity("Chengdu").withPostalCode("PANDA")
+                .withBirthday("01.01.1979").withGithubUsername("panda").withTags(td.colleagues, td.friends).build();
+        addPersonDialog.enterNewValues(pandaWong).clickCancel();
```
damithc: You may want to add a few more persons to our typical data class (but don't add them to the list). That way you can avoid copy pasting code like this many times. Remember to follow the same convention when naming persons.

src/test/java/guitests/PersonNewGuiTest.java
```diff
@@ -0,0 +1,121 @@
...
+        assertFalse(personListPanel.contains(pandaWong));
+    }
+
+    @Test
+    public void cancelOperation() {
+
+        //New
```
damithc: Redundant/incomplete comment.

src/test/java/guitests/guihandles/EditPersonDialogHandle.java
```diff
@@ -26,15 +38,27 @@
 
     public EditPersonDialogHandle(GuiRobot guiRobot, Stage primaryStage, String stageTitle) {
         super(guiRobot, primaryStage, stageTitle);
+        assertTrue(isShowingEditDialog());
     }
 
-    public boolean isValidEditDialog() {
+    /**
+     * Checks if the Edit Dialog is rendered correctly.
+     * @return
```
damithc: Don't leave empty tags.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -124,37 +128,35 @@ public void cancelOperation_usingAccelerator() {
...
-        assertNull(personListPanel.getPersonCardHandle(pandaWong));
-        personListPanel.isListMatching(td.alice, td.benson, td.charlie, td.dan, td.elizabeth);
+        assertTrue(personListPanel.isListMatching(newerAlice, td.benson, td.charlie, td.dan, td.elizabeth));
     }
 
-    //TODO: testing edits during grace period
+    @Test
+    public void cancelOperation_afterGracePeriod() {
+        //Tested in editPerson_usingContextMenu()
+    }
```
damithc: Did you include a test to confirm the TAB order is working correctly for the edit dialog?

damithc:
Added some more comments.

src/test/java/guitests/PersonEditGuiTest.java
yl-coder: Yes, I have a TODO to implement handling of multiple dialog

src/test/java/guitests/PersonEditGuiTest.java
yl-coder: My idea was to separate the button and accelerator test to its own separate classes. For example, the method here is just to test if the button open the edit dialog box. So the main thing that is testing is whether the button works but not on editing person.

And on editing person test, the test should be testing on the editing person process and not whether if the button opens the dialog box.

I guess it is a matter of definition here. I feel that it will help in preventing a bloating test class(testing too many things).

src/test/java/guitests/PersonEditGuiTest.java
yl-coder: No, should it be in the guiunittest?

yl-coder:
made changes based on the comments. I will still improve other areas of the code. But i guess i will do it on another PR. Since there is a bug fix that needs to be pushed in for other PR.

yl-coder:
feel free to comment now, but it not perfect yet. Will look through it again.

src/test/java/address/testutil/TypicalTestData.java
```diff
@@ -4,29 +4,52 @@
...
     public Person elizabeth = new Person("Elizabeth", "F. Green", 5);
-    //TODO: add more details to these persons.
+    public Person elizabethEdited = new PersonBuilder(elizabeth.copy()).withLastName("Green").build();
+    public Person fiona = new PersonBuilder("Fiona", "Wong", 6).withStreet("51th street").withCity("New York")
+                                                               .withPostalCode("51245").withBirthday("01.01.1980")
+                                                               .withGithubUsername("fiona").withTags(friends).build();
+    public Person george = new PersonBuilder("George", "David", 7).withStreet("8th Ave").withCity("Chicago")
+                                                                  .withPostalCode("25614").withBirthday("01.12.1988")
+                                                                  .withGithubUsername("george").withTags(friends, colleagues)
+                                                                  .build();
```
damithc: If these are used only for edit tests, they should not be added to typical data (which are used by all tests). They should be defined in the test itself, but in one place.

src/test/java/address/testutil/TypicalTestData.java
```diff
@@ -39,4 +62,6 @@ public TypicalTestData() {
         book.addTag(friends);
     }
 
+
+
```
damithc: Unnecessary blank lines.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -29,70 +30,80 @@ protected AddressBook getInitialData() {
...
 
+        //Confirm cancel operation does not cancel the edit after the grace period.
+        personListPanel.use_PERSON_CHANGE_CANCEL_ACCELERATOR();
         //Confirm the underlying person object has the right values
-        assertEquals(newAlice.toString(), personListPanel.getSelectedPerson().toString());
-
+        assertMatching(alicePersonCard, newAlice);
 
         //Confirm again after the next sync
         sleep(getTestingConfig().getUpdateInterval(), TimeUnit.MILLISECONDS);
```
damithc: Abstract into a 'sleepUntilNextSync' and add a TODO inside it to optimize. i.e. actively check for sync status rather than sleep a fixed time.

src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -29,70 +30,80 @@ protected AddressBook getInitialData() {
...
         assertEquals(newAlice.toString(), personListPanel.getSelectedPerson().toString());
 
         //Confirm other cards are unaffected.
-        personListPanel.clickOnListView();
         assertTrue(personListPanel.isListMatching(1, td.benson, td.charlie, td.dan, td.elizabeth));
 
         //Confirm status bar is updated correctly
-        assertEquals(statusBar.getText(), "Edit Person [ Alice Brown -> Alicia Brownstone ] completed successfully.");
+        assertEquals(HeaderStatusBarHandle.formatSuccessMessage(HeaderStatusBarHandle.Type.EDIT, td.alice.fullName(),
+                     Optional.of(newAlice.fullName())), statusBar.getText());
```
damithc: Still showing too much details. abstract more e.g. `HeaderStatusBarHandle.getEditCanceledMessage(td.alice)`

src/test/java/guitests/PersonEditGuiTest.java
damithc: Sure.

src/test/java/guitests/PersonNewGuiTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Test
+    public void addPerson_cancelDialog() {
+        EditPersonDialogHandle addPersonDialog = personListPanel.clickNew();
+        addPersonDialog.enterNewValues(td.fiona).clickCancel();
+        assertFalse(personListPanel.contains(td.fiona));
+    }
+
+    @Test
+    public void cancelOperation() {
```
damithc: Why is this named differently?

src/test/java/guitests/PersonNewGuiTest.java
```diff
@@ -0,0 +1,94 @@
...
+ */
+public class PersonNewGuiTest extends GuiTestBase {
+
+    @Override
+    protected AddressBook getInitialData() {
+        return td.book;
+    }
+
+    @Test
+    public void addPerson_createAndDelete() {
```
damithc: `createAndDeleteInGracePeriod`?

src/test/java/guitests/PersonNewGuiTest.java
```diff
@@ -0,0 +1,94 @@
...
+        assertFalse(personListPanel.contains(td.fiona));
+
+        //Add George
+        addPersonDialog = personListPanel.clickNew();
+        addPersonDialog.enterNewValues(td.george).clickOk();
+        PersonCardHandle georgeCard = personListPanel.navigateToPerson(td.george);
+        assertTrue(georgeCard.isShowingGracePeriod("Adding"));
+        sleepForGracePeriod();
+        assertFalse(personListPanel.contains(td.fiona)); //Make sure Person A doesn't appear again.
+        assertTrue(personListPanel.contains(td.george));
```
damithc: Why not do a full list match here?

src/test/java/guitests/PersonNewGuiTest.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Test
+    public void addPerson_multipleAdd() {
```
damithc: What do you mean by `multipleAdd`? Be more specific.

yl-coder:
Merging this one. Will continue in another PR.

src/test/java/guitests/PersonEditGuiTest.java
yl-coder: If without the type. We need to have 3 different type of similar methods.
getEditCancelledMessage(), getAddCancelledMessage(), getDeleteCancelledmessage.

And "get.." seems to suggesting it is getting something from the statusBar. 

Messages may need two different names. For e.g. for this case [ Alice Brown -> Alicia Brownstone ].

All these cases complicate the convenience method. I guess showing the plain text will be more appropriate?

src/test/java/guitests/PersonNewGuiTest.java
yl-coder: Earlier, you commented this. Therefore i changed from cancelPerson -> cancelOperation.

I guess i can change it to addPerson_cancelOperation.

yl-coder: or addPerson_cancelAddOperation

PR #564, opened 2016-07-26
Add full GUI testing for tagging #559

### LINKED ISSUE #559: Add full GUI testing for tagging
Test both the accelerator method and within the person edit dialog.

m133225:
Fixes #559


src/test/java/guitests/PersonEditGuiTest.java
```diff
@@ -156,5 +157,19 @@ public void cancelOperation_usingAccelerator() {
...
+        EditPersonDialogHandle bensonEditDialog = personListPanel.use_PERSON_EDIT_ACCELERATOR();
+        bensonEditDialog.focusOnWindow(EDIT_TITLE);
+        TagPersonDialogHandle bensonTagDialog = bensonEditDialog.openTagPersonDialogUsingShortcut();
+        bensonTagDialog.enterSearchQuery("coll").acceptSuggestedTag();
+        bensonTagDialog.close();
+        bensonEditDialog.pressEnter();
+        assertEquals("Editing", bensonPersonCard.getPendingStateLabel());
+        sleepForGracePeriod();
+        assertEquals("Tag: colleagues", bensonPersonCard.getTags());
+    }
```
damithc: This should be a separate class altogether?

src/test/java/guitests/TagPersonGuiTest.java
```diff
@@ -0,0 +1,276 @@
...
+        assertEquals("Tag: colleagues", alicePersonCard.getTags());
+        assertTrue(alicePersonCard.isShowingGracePeriod("Editing"));
+        sleepForGracePeriod();
+        assertEquals("Tag: colleagues", alicePersonCard.getTags());
+        assertFalse(alicePersonCard.isShowingGracePeriod("Editing"));
+
+        EditPersonDialogHandle aliceEditDialogTwo = personListPanel.use_PERSON_EDIT_ACCELERATOR();
+        TagPersonDialogHandle aliceTagDialogTwo = aliceEditDialogTwo.openTagPersonDialogUsingShortcut();
+
+        aliceTagDialogTwo.enterSearchQuery("coll").acceptSuggestedTag();
```
damithc: The two calls are used together frequently. Can be combined to one, and chained.
`aliceTagDialog.addTag("coll").addTag("Fri")...`

src/test/java/guitests/TagPersonGuiTest.java
```diff
@@ -0,0 +1,276 @@
...
+        aliceTagDialog.enterSearchQuery("coll").acceptSuggestedTag();
+        aliceTagDialog.close();
+
+        aliceEditDialog.pressEnter();
+
+        assertEquals("Tag: colleagues", alicePersonCard.getTags());
+        assertTrue(alicePersonCard.isShowingGracePeriod("Editing"));
+        sleepForGracePeriod();
+        assertEquals("Tag: colleagues", alicePersonCard.getTags());
+        assertFalse(alicePersonCard.isShowingGracePeriod("Editing"));
```
damithc: Write a convenience method (in this class itself) to do these 5 steps? It can be used in many other places in this class.

src/test/java/guitests/TagPersonGuiTest.java
```diff
@@ -0,0 +1,276 @@
...
+import static org.junit.Assert.assertTrue;
+
+public class TagPersonGuiTest extends GuiTestBase {
+    @Override
+    protected AddressBook getInitialData() {
+        return td.book;
+    }
+
+    @Test
+    public void tagPerson() {
```
damithc: Various cases of adding one tag at a time.
1. Add to a person without tags.
2. Add to a person with existing tags.
3. Add tag to a person having the same tag already.

damithc:
Added a few comments in passing. No a full review. Try to reduce verbosity of code by chaining and defining convenience methods (in place of repeated code blocks) in PageObject or the Test class itself.

m133225:
Seems to pass locally consistently. Will wait until #557 to further debug. 

Also, when selecting multiple persons, each card handle tries to focus itself, causing large amounts of processing.

damithc:
I think there is an undocumented keyboard shortcut in the edit dialog (Ctrl+O) that is not tested anywhere. Perhaps it's best to remove it? If not, we need to document it, test it, and also figure out how it should fit in with our keybinding component, which is supposed to capture all key bindings.

m133225:
It is in some of the tests already. However, I think it is better we remove it for now. Maybe we can introduce it back in future versions so as to implement bindings in dialogs (instead of just the main window), which could be useful for HT.

m133225:
Tests for tagging multiple persons seem to fail at verifying (on Travis) before grace period ends. Will look into it.

yl-coder:
There is a typo in travis yml which causes screenshots not being able to upload.
rebase to latest master to see screenshots

m133225:
Selecting multiple persons seem to not be working in headless mode, not just on Travis. Will attempt to run them in headful mode instead.

PR #566, opened 2016-07-27
Add full GUI tests for filters #560

### LINKED ISSUE #560: Add full GUI tests for filters
As per title

m133225:
Fixes #560

### DIFF
--- a/src/test/java/guitests/FilterPersonsGuiTest.java
+++ b/src/test/java/guitests/FilterPersonsGuiTest.java
@@ -11,7 +11,6 @@ import org.junit.Test;
 import java.time.LocalDate;
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.concurrent.TimeUnit;
 
 import static org.junit.Assert.assertTrue;
 
--- a/src/test/java/guitests/FilterPersonsGuiTest.java
+++ b/src/test/java/guitests/FilterPersonsGuiTest.java
@@ -3,11 +3,15 @@ package guitests;
 import address.model.datatypes.AddressBook;
 import address.model.datatypes.person.Person;
 import address.model.datatypes.tag.Tag;
+import commons.StringUtil;
 import org.junit.Test;
 
 import java.time.LocalDate;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
 
 import static org.junit.Assert.assertTrue;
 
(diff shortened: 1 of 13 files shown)


src/test/java/guitests/FilterPersonsGuiTest.java
```diff
@@ -0,0 +1,119 @@
...
+    public AddressBook getInitialData() {
+        john = new Person("John", "Tan", 1);
+        mary = new Person("Mary", "Jones", 2);
+        annabelle = new Person("Annabelle", "Muellers", 3);
+        barney = new Person("Barney", "Rogers", 4);
+        charlie = new Person("Charlie", "Chip", 5);
+        danny = new Person("Danny", "Rogers", 6);
+        colleagues = new Tag("colleagues");
+        friends = new Tag("friends");
+        family = new Tag("family");
```
damithc: Why not use the typical data class? That one is supposed to have a well-defined set of data to be used for testing.

src/test/java/guitests/guihandles/PersonListPanelHandle.java
```diff
@@ -394,6 +395,20 @@ public boolean isListMatching(int startPosition, Person... persons) {
         return true;
     }
 
+    public boolean isExactList(Person... persons) {
```
damithc: I think @yl-coder defined a similar method in his ongoing PR.

src/test/java/guitests/FilterPersonsGuiTest.java
m133225: Will add to that data. The current set of data lacks tags being assigned to people and the fields are all unique (can't get filter results containing several people e.g. no 2 people with name `Muellers`).

src/test/java/guitests/guihandles/PersonListPanelHandle.java
yl-coder: Yes. But the implementation of this method is different from mine.  I think for guitest, we should not be grabbing the underlying list from the listview. Comparision should be done between what is shown on the listview by grabbing values from the cardhandle.

src/test/java/guitests/FilterPersonsGuiTest.java
m133225: Just noticed, if we are to follow the current name pattern, it will be difficult to have 2 people with the same (partial) name.

src/test/java/guitests/guihandles/PersonListPanelHandle.java
m133225: Will use your implementation when your PR is merged. Will use this in the meantime as a workaround.

src/test/java/guitests/FilterPersonsGuiTest.java
damithc: You add those persons within your test. i.e. take the typical data set and configure it to fit your test. Any data you think that will be useful for others can be added to the common data set. e.g. you can add a `td.personWithSameFirstNameAsAlice`

m133225: Hmm, yes, but I think having duplicate names may be using for the common data set too.

m133225: Think there will be alot of conflicts if I change this test data though. Will probably need to wait till the other PR is done.

m133225:
@yl-coder Note that I reused some of your edited persons as typical test data in this PR.

PR #568, opened 2016-07-27
Improve and add GUI test cases part 2

yl-coder:
Continuation from #557

### DIFF
--- a/src/test/java/address/testutil/ContextMenuChoice.java
+++ b/src/test/java/address/testutil/ContextMenuChoice.java
@@ -1,0 +1,8 @@
+package address.testutil;
+
+/**
+ *
+ */
+public enum ContextMenuChoice {
+        EDIT, TAG, DELETE, CANCEL;
+}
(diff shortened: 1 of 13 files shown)


src/test/java/address/testutil/ContextMenuChoice.java
```diff
@@ -0,0 +1,8 @@
+package address.testutil;
+
+/**
+ *
+ */
+public enum ContextMenuChoice {
+        EDIT, TAG, DELETE, CANCEL;
+}
```
damithc: Can be define this in another class? Doesn't look like it should be on its own, at least based on current class name.

damithc: May be in the PersonListGuiHandle?

yl-coder: Maybe i will remove this. Will have four different methods instead of one unified method

damithc: That's not going to be scalable though. A context menu can grow to have many options.

yl-coder: yea, in anyways tester still need to modified the existing codes when there are more options. Having small methods rather than the current one big methods help in the readability also.

PR #578, opened 2016-07-29
Resolve headless testing disruption for Mac users #575

### LINKED ISSUE #575: Resolve headless testing disruption for Mac users
Either by creating a custom task without the problematic tests or by identifying and resolving the problematic component.

m133225:
Fixes #575


damithc:
We should provide a way for mac users to omit those tests if they wish to, but should not omit them by default.

PR #579, opened 2016-07-29
Add gui unit test for edit dialog

yl-coder:
As per title


damithc:
@m133225 can you review?

src/test/java/address/testutil/ApplicationTest.java
```diff
@@ -0,0 +1,80 @@
...
+    // METHODS.
+    //---------------------------------------------------------------------------------------------
+
+    @Before
+    @Unstable(reason = "is missing apidocs")
+    public final void internalBefore()
+            throws Exception {
+        FxToolkit.registerPrimaryStage();
+        FxToolkit.setupApplication(() -> new ApplicationAdapter(this));
+        sleep(1000);
```
m133225: Is this class simply to add the sleep duration? I think we can extend `ApplicationTest` instead?

src/test/java/address/testutil/ApplicationTest.java
```diff
@@ -0,0 +1,80 @@
...
+import org.junit.Before;
+import org.testfx.api.FxToolkit;
+import org.testfx.api.annotation.Unstable;
+import org.testfx.framework.junit.ApplicationAdapter;
+import org.testfx.framework.junit.ApplicationFixture;
+
+/**
+ * Code mostly from TestFX
+ */
+public abstract class ApplicationTest extends GuiRobot implements ApplicationFixture {
```
m133225: Instead of this class, we can create something similar to `GuiTestBase` e.g. `GuiUnitTestBase` which implements `ApplicationFixture` for the `start`, `stop` and `init` methods.

This is because we do not need the base class to extend `GuiRobot`/`FxRobot` any more, and we can directly use the `handle` variable in the test class.

src/test/java/guiunittests/GuiUnitTestBase.java
```diff
@@ -0,0 +1,66 @@
...
+    public final void before()
+            throws Exception {
+        FxToolkit.registerPrimaryStage();
+        FxToolkit.setupApplication(() -> new ApplicationAdapter(this));
+        guiRobot.sleep(1000);
+    }
+
+    @After
+    public final void after()
+            throws Exception {
```
damithc: Why line break?

src/test/java/guiunittests/GuiUnitTestBase.java
```diff
@@ -0,0 +1,66 @@
...
+import org.testfx.api.FxToolkit;
+import org.testfx.api.annotation.Unstable;
+import org.testfx.framework.junit.ApplicationAdapter;
+import org.testfx.framework.junit.ApplicationFixture;
+
+import java.util.concurrent.TimeoutException;
+
+/**
+ *
+ */
```
damithc: Add comment

src/test/java/guiunittests/PersonEditDialogUnitTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+import java.io.IOException;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * To test the basic features of the edit dialog
+ */
+public class PersonEditDialogUnitTest extends GuiUnitTestBase {
```
damithc: Is this the only guiUnitTest we have? If not, others need to extend from the same parent class.

src/test/java/guiunittests/PersonEditDialogUnitTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+    TypicalTestData data = new TypicalTestData();
```
damithc: Is this being used?

damithc:
No more comments from me.

